PR #49, opened 2015-10-09
Add a broken test to simulate libinputs sys-path and device path validation

APokorny:
When libinput is asked to open a device via path it opens the device
directly but also tries to achieve the same by evaluating major minor
from fstat(fd).st_rdev and opening the resulting
/sys/dev/type/major:mior path. The information returned by fstat does
currently not match the stat data received directly from the path.

I tried to get this working without needing the file, but failed to setup the umockdev device properly.
With the given test it bails out, just like libinput would:

ERROR:tests/test-umockdev.c:302:t_testbed_add_devicev_fstat: assertion failed (stat_from_stat.st_rdev == stat_from_fstat.st_rdev): (3398 == 34850)


masterkorp:
:+1:

PR #52, opened 2016-04-21
include sys/sysmacros.h for major/minor/makedev

vapier:
These funcs are defined in the sys/sysmacros.h header, not sys/types.h.
Linux C libraries are updating to drop the implicit include, so we need
to include it explicitly.


vapier:
i'm not 100% sure this is correct as i don't have vala tools.  but it seems right :).

martinpitt:
Builds fine, thanks!

PR #65, opened 2017-12-15
thunderbolt: Create symlinks the same way like usb and pci

finikorg:
This should create /sys/bus/thunderbolt and links.


martinpitt:
Thanks! This looks good, but somehow CI croaks. I don't think this is related to this PR, though.

finikorg:
This is the error:
("Cannot call udevadm: udevadm exited with code 512\nsyspath not found\n\n" == "")

martinpitt:
@finikorg: Right, I noticed. This test case is rather demanding, I've also seen it fail on various Debian buildds where some things in /sys just don't like to be read :-)

Issue #69, opened 2018-01-20
umockdev breaks due to systemd changes


gicmo:
As seen in https://github.com/hughsie/fwupd/issues/370 (and also [boltd ci](https://travis-ci.org/gicmo/bolt/jobs/330918187)).
```
$ LD_PRELOAD='libumockdev-preload.so.0' build/plugins/thunderbolt/thunderbolt-self-test -p /thunderbolt/basic

/thunderbolt/basic: 
ERROR: uevent_sender_send: No such device /sys/devices/domain0
ERROR: uevent_sender_send: No such device /sys/devices/domain0/0-0
```
When stracing you see a change in the syscall pattern:

**working** (systemd-234-9.fc27):
```
stat("/tmp", {st_mode=S_IFDIR|S_ISVTX|0777, st_size=1400, ...}) = 0
access("/tmp/umockdev.TU2FDZ", F_OK)    = 0
stat("/tmp/umockdev.TU2FDZ", {st_mode=S_IFDIR|0700, st_size=80, ...}) = 0
access("/tmp/umockdev.TU2FDZ/sys", F_OK) = 0
stat("/tmp/umockdev.TU2FDZ/sys", {st_mode=S_IFDIR|0755, st_size=40, ...}) = 0
access("/tmp/umockdev.TU2FDZ/sys/devices", F_OK) = -1 ENOENT (No such file or directory)
mkdir("/tmp/umockdev.TU2FDZ/sys/devices", 0755) = 0
access("/tmp/umockdev.TU2FDZ/sys/devices/domain0", F_OK) = -1 ENOENT (No such file or directory)
mkdir("/tmp/umockdev.TU2FDZ/sys/devices/domain0", 0755) = 0
access("/tmp", F_OK)                    = 0
stat("/tmp", {st_mode=S_IFDIR|S_ISVTX|0777, st_size=1400, ...}) = 0
access("/tmp/umockdev.TU2FDZ", F_OK)    = 0
stat("/tmp/umockdev.TU2FDZ", {st_mode=S_IFDIR|0700, st_size=80, ...}) = 0
access("/tmp/umockdev.TU2FDZ/sys", F_OK) = 0
stat("/tmp/umockdev.TU2FDZ/sys", {st_mode=S_IFDIR|0755, st_size=60, ...}) = 0
access("/tmp/umockdev.TU2FDZ/sys/class", F_OK) = -1 ENOENT (No such file or directory)
[...]

PR #70, opened 2018-01-21
Wrap openat{,64} to accommodate for libudev changes

gicmo:
A quick proof-of-concept to mitigate issue #69 
It unbreaks the fwupd thunderbolt test for me by trapping the relevant openat64 call:

```
testbed wrapped open64(/) -> /
testbed wrapped openat64 (sys) -> /tmp/umockdev.9S6CDZ/sys
testbed wrapped openat64 (devices) -> devices
testbed wrapped openat64 (domain0) -> domain0
testbed wrapped openat64 (0-0) -> 0-0
testbed wrapped openat64 (0-1) -> 0-1
testbed wrapped openat64 (0-2) -> 0-2
```
This is just WIP and might need cleanup.

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -1199,6 +1199,47 @@ WRAP_OPEN(, 64);
 WRAP_OPEN2(__,_2);
 WRAP_OPEN2(__,64_2);
 
+/* wrapper template for openat family; intercept opening /sys from the root dir */
+#define WRAP_OPENAT(prefix, suffix) \
+int prefix ## openat ## suffix (int dirfd, const char *pathname, int flags, ...)		\
+{ \
+    const char *p = NULL;									\
+    libc_func(prefix ## openat ## suffix, int, int, const char *, int, ...);			\
+    libc_func(readlink, ssize_t, const char*, char *, size_t);					\
+    int trapped = 0, ret;									\
+    TRAP_PATH_LOCK;										\
+  \
+    if (strncmp(pathname, "sys", 3) == 0 && (pathname[3] == '/' || pathname[3] == '\0')) {	\
+        static char buf[PATH_MAX],link[PATH_MAX];						\
+        snprintf(buf, sizeof(buf), "/proc/self/fd/%d", dirfd);					\
+        if (_readlink(buf, link, sizeof(link)) == 1 && link[0] == '/') {			\
+            trapped = 1;									\
+            strncpy(link + 1, pathname, sizeof(link) - 1);					\
+            buf[sizeof(link) - 1] = 0;								\
+            p = trap_path(link);								\
+        } \
+    } \
+  \
+    if (!trapped)										\
+        p = trap_path(pathname);								\
+    DBG(DBG_PATH, "testbed wrapped " #suffix "openat" #suffix "(%s) -> %s\n", pathname, p);	\
+    if (p == NULL) { TRAP_PATH_UNLOCK; return -1; }						\
+    if (flags & (O_CREAT | O_TMPFILE)) {							\
+	mode_t mode;										\
+	va_list ap;										\
+	va_start(ap, flags);									\
+	mode = va_arg(ap, mode_t);								\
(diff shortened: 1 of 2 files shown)


martinpitt:
Summary from yesterday's IRC conversation:
 * The approach of reading /proc/fd is a bit ugly, but I don't see a better way either. Limiting the resolution to the root directory should be okay though, as long as udev is satisfied with that.
 * Please add a small test case to make sure this works as expected, also on other architectures. Particularly on i386, where you might need a variation of this (without the 64 suffix). I don't currently run a CI test on 32 bit, but I should add that :-)
 * Would be nice to not duplicate so much code, but reuse the OPEN* macros and/or factorize some of their logic. But this isn't very important, feel free to leave as-is.

xnox:
New systemd/udev landed in bionic-proposed, I can confirm that it fails on all arches =)

```
autopkgtest for umockdev/0.10-1: amd64: Regression ♻ , arm64: Test in progress, armhf: Regression ♻ , i386: Regression ♻ , ppc64el: Regression ♻ , s390x: Test in progress
```

I want to see if I can fix this up across the board.

xnox:
Applying this patch, and running the autopkgtests on both amd64 and i386 passes just fine. Either openat64 is used on i386, and/or the i386 code path remained the same.... Will take this patch and upload to Ubuntu to test all architectures.

xnox:
Not sure how to add a test case for this. None of the other overwriten functions are unit-tested, are they?

martinpitt:
@xnox: tests/test-umockdev.c covers these; whether you call these unit or integration tests is up to you :-) But if the tests now  fail with the new libudev 236, then they actually do cover this. I wouldn't mind a more specific `openat()` test for the case that the directory is not NULL or `/`, as libudev doesn't cover that; but it's not too important indeed.

gicmo:
@xnox I had this on my TODO and wanted too look at this tomorrow, but if you want to take over, I would not mind. ;)

gicmo:
@xnox you working on this? Otherwise I can take a look at it today.

xnox:
@gicmo well, I didn't do much, so do continue with this. This patch alone, has the following results:

```
autopkgtest for umockdev/0.10-1ubuntu1: amd64: Pass, arm64: Regression ♻ , armhf: Regression ♻ , i386: Pass, ppc64el: Pass, s390x: Pass
```

So for some reason arm* is special, investigating

martinpitt:
I took this opportunity to finally set up a better CI in #71 and #72 to not just run tests on the ancient Ubuntu 14.04 (what travis gives you), but on xenial, artful, and current devel (bionic at this time). This still does not reproduce the bug (I don't want CI to fail in master), but I added support for a `$PROPOSED` test env var parameter, which nicely reproduces this issue and confirms the fix:

    PROPOSED=1 tests/run-ubuntu-chroot

in a clean checkout; this defaults to the current Ubuntu devel series, i. e. bionic. I don't want to run that in CI as that would be too unreliable, but at least it's a convenient way to try locally.

martinpitt:
@xnox: looks like the autopkgtest worked now. Seems on arm there is some race condition in `t_testbed_script_replay_socket_stream`, but that's unrelated to the openat issue.

superm1:
@martinpitt  any reason to prefer chroots over docker containers?

martinpitt:
@gicmo : I force-pushed your branch to rebase to current master (for the recent tests) and add a few small robustifications and simplifications. Still working on verifying on 32 bits and adding tests.

@superm1: I found this to be a lot simpler and faster, and I could just steal the script from python-dbusmock with just minimal changes.

martinpitt:
@gicmo: I now wrote some tests which both directly show the regression with calling `openat64()`, as well as show that even on a 64 bit machine `openat()` is not the same as `openat64()`, so both need to be wrapped. I did that now by turning this into a macro like for the open and stat families. There's still some duplication, but this isn't trivial to untangle and it does not need to block this any further. Thanks for your nice work here!

PR #86, opened 2018-09-22
Support CentOS 7 in Travis CI

mrc0mmand:
I played around a little with Docker in Travis and here are the results. First of all - this is going to fail until #85 is merged, then I'll rebase this branch. I'm submitting this PR beforehand, as I'd like to discuss one issue I come across during the testing.

The test suite appears to be passing during the build in Fedora Copr, as I forgot to add the `python-gobject` package to the dependencies. Thus, the `make check` just skips all the test using the Python GI:

```
== Running GI test with python against build tree ==
GI module not available, skipping test: No module named gi
```

After I installed the missing package, the `make check` started failing:
```
======================================================================
ERROR: test_add_from_string_errors (__main__.Testbed)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/test-umockdev.py", line 194, in test_add_from_string_errors
    with self.assertRaisesRegex(GLib.GError, 'must start with.*P:') as cm:
AttributeError: 'Testbed' object has no attribute 'assertRaisesRegex'
```

According to the [Python 3 documentation](https://docs.python.org/3.3/library/unittest.html#unittest.TestCase.assertRaisesRegex) the `assertRaisesRegex` method used to be called `assertRaisesRegexp` before Python 3.2. CentOS doesn't have Python 3 installed by default, which results in the aforementioned fail.

I think there are two possible solutions - either to install Python 3 from the EPEL or to slightly modifiy the test suite. I decided to do the latter, which can be found in the commit e65e99f. Of course, that's just my idea - if the Python 3 installation would be a better choice, I can drop the offending commit.

For testing purposes I merged patches from this PR with the patch from the #85 to see if everything works as expected - and it does, see https://travis-ci.org/mrc0mmand/umockdev/builds/431913447

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
 - TEST=tests/run-ubuntu-chroot RELEASE=bionic ARCH=i386
 - TEST=tests/run-ubuntu-chroot RELEASE=xenial
 - TEST=tests/run-alpine-docker RELEASE=latest
+- TEST=tests/run-centos-docker RELEASE=7
 script: $TEST
 cache:
   directories:
--- a/tests/run-centos-docker
+++ b/tests/run-centos-docker
@@ -1,0 +1,24 @@
+#!/bin/sh
+set -eu
+
+# keep container around if $DEBUG is set, and allow strace
+RM=""
+[ -n "${DEBUG:-}" ] && OPTS="--privileged" || OPTS="--rm"
+
+sudo docker run --interactive $OPTS --volume `pwd`:/source centos:${RELEASE:-latest} /bin/sh << EOF
+
+# install build dependencies
+yum -y install autoconf automake gcc glib2-devel gtk-doc libgudev1-devel libtool libudev-devel make python-gobject vala
+
+# run build as user
+useradd guest
+su -s /bin/sh - guest << EOG
+set -ex
+cp -r /source /tmp
+cd /tmp/source
+./autogen.sh
+make -j4
+make check
+EOG
+
+EOF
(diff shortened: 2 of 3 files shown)


martinpitt:
This looks good indeed, thanks! Personally I prefer a more robust EAFP style like

```py
try:
    assertRaisesRegex = self.assertRaisesRegex
except AttributeError:
    assertRaisesRegex = self.assertRaisesRegexp  # py 2 and old py 3
```
that avoids harcdoding assumptions about version numbers. That's of course bikeshedding and matter of preference.

There are still some test skips like

    /umockdev-testbed-usb/lsusb: SKIP: lsusb not installed. OK
    /umockdev-run/integration/input-touchpad: [SKIP: Xorg not installed] OK
    /umockdev-run/integration/input-evtest-evemu: which: no evtest in (./src:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin)
    [SKIP: evtest not installed] OK

As these are the more elaborate parts of umockdev that are good for finding bugs, it would be nice to install the packages for these as well. But if these cause any failures, it's totally fine to enable them step by step and fix things in follow-ups. Even with a few skips, running the build and most tests is already highly beneficial and avoids regressions.

So, can you please rebase to current master to pick up the `O_TMPFILE` fix, possibly update the RaisesRegexp check, and try how far you get with lsusb and evdev? Thanks!

mrc0mmand:
The EAFP style is definitely more robust, thanks.

The situation with dependencies is a little more complicated. So far, there were missing following dependencies:
* evtest
* gphoto2
* lsusb
* Xorg

The only dependencies currently available in CentOS/EPEL repositories are lsusb (package usbutils), and Xorg. So far I added at least the usbutils package to the CI and thankfully it works. As for the Xorg related tests - even with installed dependencies the Xorg fails to start, yet the error log is empty, so I guess I'll need to dig a little bit deeper to fix this one.

PR #90, opened 2019-06-25
Implement TODO for devnode symlinks

dlech:
This implements creating symlinks for S: lines in umockdev device files.

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -956,6 +956,7 @@ public class Testbed: GLib.Object {
         string cur_data = data;
         string? devnode_path = null;
         uint8[] devnode_contents = {};
+        string[] devnode_links = {};
 
         cur_data = this.record_parse_line(cur_data, out type, out key, out devpath);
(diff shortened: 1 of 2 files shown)


src/umockdev.vala
```diff
@@ -1018,7 +1019,10 @@ public class Testbed: GLib.Object {
                     break;
 
                 case 'S':
-                    /* TODO: ignored for now */
+                    /* collect symlinks */
+                    if (val != null) {
+                        devnode_links += Path.build_filename(this.root_dir, "dev", val);
```
martinpitt: An `S:` symlink should always have a value, otherwise the line is pointless. Let's error out for empty values (`throw UMockdev.Error.PARSE`) instead of silently ignoring it.

martinpitt:
I fixed the [failing tests](https://travis-ci.org/martinpitt/umockdev/jobs/550547525) in master now. If you could rebase your PR on top of current master, that'd be great!

dlech:
updated

PR #93, opened 2019-08-18
Don't expect size to be non-zero for mocked /sys

abbradar:
This isn't true for tmpfs, which leads to hard to debug errors.

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -1740,7 +1740,7 @@ public bool in_mock_environment()
     if (!__in_mock_env_initialized) {
         Posix.Stat st;
         if (Posix.stat("/sys", out st) >= 0)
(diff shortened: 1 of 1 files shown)


abbradar:
Yeah, it actually was btrfs or ZFS, not tmpfs. 

On August 28, 2019 9:25:32 AM GMT+02:00, Martin Pitt <notifications@github.com> wrote:
>martinpitt approved this pull request.
>
>I run on tmpfs all the time and haven't noticed any fallout. However,
>this does make sense, thanks!
>
>I just fixed the ubuntu devel test failure on master, that was
>unrelated.
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/martinpitt/umockdev/pull/93#pullrequestreview-280634687

-- 
Nikolay.

PR #94, opened 2019-08-28
Add all bus names from Linux kernel 5.3

dlech:
This adds a lookup matching function to match all (current) bus names
isntead of just 3 names.


martinpitt:
Indeed this is super-super-ugly and will require constant maintenance :sob: But indeed I don't have any other great idea either -- umockdev-record is not a required tool for device emulation to work, test suites are fully entitled to construct the devices directly. That should probably specify explicitly if it's a bus or a class, but that makes the API harder to use and also backwards incompatible. So let's go with that for now, thank you!

Just restarted the two failed tests; I fixed CI a few days ago.

martinpitt:
I force-pushed your branch to rebase to current master and fix CI.

PR #103, opened 2020-07-28
Fix flaky test operations

ddstreet:
this includes several commits to stop some of the flakiest test cases that cause lots of autopkgtest false failures on armhf (and occasionally on other archs)

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -1618,12 +1618,12 @@ private class ScriptRunner {
     {
         Posix.fd_set fds;
         Posix.timeval timeout = {0, 200000};
-        size_t offset = 0;
+        ssize_t len = 0;
         uint8[] buf = new uint8 [data.length];
 
         // a recorded block might be actually written in multiple smaller
         // chunks
-        while (this.running && offset < data.length) {
+        while (this.running && len < data.length) {
             Posix.FD_ZERO (out fds);
             Posix.FD_SET (this.fd, ref fds);
             int res = Posix.select (this.fd + 1, &fds, null, null, timeout);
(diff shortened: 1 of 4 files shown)


tests/test-umockdev-record.vala
```diff
@@ -386,7 +386,6 @@ t_system_script_log_simple ()
     assert_cmpuint (logwords.length, Op.EQ, 3);
     assert_cmpstr (logwords[0], Op.EQ, "r");
     // should be quick, give it 5 ms at most
-    assert_cmpint (int.parse(logwords[1]), Op.LE, 5);
```
martinpitt, line 389: It's fine to skip the tests on ARM with `#ifdef` and such, or multiply all times (script and timeout) by some factor. But I'd like to keep them at least for x86. The whole point of the test is to validate that the preload library actually implements the correct timing of the script.

src/umockdev.vala
```diff
@@ -1636,43 +1636,48 @@ private class ScriptRunner {
 
             if (res == 0) {
                 debug ("ScriptRunner[%s]: timed out on read operation on expected block '%s', trying again",
-                       this.device, encode(data[offset:data.length]));
+                       this.device, encode(data[len:data.length]));
                 continue;
             }
 
-            ssize_t len = Posix.read (this.fd, buf, data.length - offset);
+            ssize_t ret = Posix.read (this.fd, &buf[len], data.length - len);
```
martinpitt, line 1643: Ouch! Well spotted.

tests/test-umockdev.c
```diff
@@ -1821,8 +1821,9 @@ assert_delta_t(const struct timeval * first, const struct timeval * second, int
 {
     int delta = (second->tv_sec - first->tv_sec) * 1000;
     delta += (second->tv_usec - first->tv_usec) / 1000;
-    g_assert_cmpint(delta, >=, ms - 50);
-    g_assert_cmpint(delta, <=, ms + 50);
+    /* specific timing can never be guaranteed if the testbed is not controlled */
+    if (delta < ms - 50 || delta > ms + 50)
+        g_printf("Warning: delta %dms out of expected %dms -/+ 50ms range.", delta, ms);
```
martinpitt, line 1826: Same here -- if the test doesn't fail, it's pointless to run it in the first place. Let's rather skip it on slow architectures, or apply a factor.

martinpitt:
It seems Travis is AWOL -- I need to look at this first before anything can be landed.

ddstreet:
> Thanks!
> 
> I haven't read through the entire PR yet, just some initial notes.
> 
> I just wonder where this happens? [Debian armhf builds](https://buildd.debian.org/status/logs.php?pkg=umockdev&arch=armhf) have a 100% perfect history there, the bunch of failures were not related to timing test failures. I looked at [Ubuntu's builds](https://launchpad.net/ubuntu/bionic/armhf/+builds?build_text=umockdev&build_state=failed) for a few releases, and didn't find any failures.

Ubuntu doesn't have many builds since it's just synced from Debian in almost all cases, e.g.:
https://launchpad.net/ubuntu/bionic/armhf/+builds?build_text=umockdev&build_state=built
https://launchpad.net/ubuntu/focal/armhf/+builds?build_text=umockdev&build_state=built

The autopkgtests are really where things fall apart; armhf is by far the worst:
http://autopkgtest.ubuntu.com/packages/umockdev/bionic/armhf
http://autopkgtest.ubuntu.com/packages/umockdev/focal/armhf
http://autopkgtest.ubuntu.com/packages/umockdev/groovy/armhf

but it's not limited to only armhf; there are less frequent false negatives on other archs, too, e.g.:
http://autopkgtest.ubuntu.com/packages/umockdev/eoan/i386
http://autopkgtest.ubuntu.com/packages/umockdev/eoan/arm64
http://autopkgtest.ubuntu.com/packages/umockdev/bionic/ppc64el

I don't have direct access to the autopkgtest infrastructure, so you probably know more about it than me, but my understanding is armhf is run inside an armhf LXD container which is running on a arm64 openstack instance. I did try to manually reproduce the timing failures by doing the same, i.e. creating a arm64 instance with an armhf container and running the test in a loop there, but i couldn't reprdouce any timing failure; it's possible though that the autopkgtest infrastructure is slower and/or highly overloaded.

tests/test-umockdev-record.vala
ddstreet: i think the problem is this strict time limits just aren't possible to achieve without controlling the testbed. the ubuntu autopkgtests, by nature of running in an openstack cloud, simply can't guarantee that *any* action will happen within a 5ms window.

I don't think #ifdef'ing for armhf would really be appropriate, since it's possible the problem is that armhf is slow on the ubuntu autopkgtest systems; but I don't know if that means it's slow on *all* armhf systems.

Probably the best approach then if you want to keep it is to have a minimum-time multipler as you mentioned, i'll see if i can adjust it.

ddstreet:
ok i force pushed some updates, adding a timing factor variable, i'm building a package to run in the ubuntu autopkgtest armhf system repeatedly now.

martinpitt:
Thanks @ddstreet ! That makes sense. How about I rework the autopkgtests to not run any timing related test at all? IMHO these are most important for local development, and as they seem to work well enough during package builds I'd also like to keep them there, to make sure that the code works on all architectures. But gating reverse dependencies has a different purpose, and it'd be sufficient to only run a subset of robust tests there, making sure that the library and CLI programs work in general. WDYT?

martinpitt:
The code changes look great now, thanks! But I still need to fix Travis so that tests run.

martinpitt:
Travis [did run](https://travis-ci.org/github/martinpitt/umockdev/builds/712706107), but for some reason fails to report back statuses to GitHub. Apparently this is fallout from the [travis-ci.org → travis-ci.com migration](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/). But I've tried this for like 30 minutes now, and it just refuses to cooperate..

martinpitt:
Released: https://github.com/martinpitt/umockdev/releases/tag/0.14.2

I upload this into Debian now. For the autopkgtests I'll set the slow factor to 10, that's enough as a smoke test. Upstream CI and local unit tests are still strict then.

martinpitt:
.. or I don't upload it yet, as newer gcc/autotools/etc. makes `dh_autoreconf` fail. Investigating..

PR #107, opened 2020-10-22
preload: Better debug for functions missing from libc

hadess:
If finding one of the symbols for a function fails, whichever function
it is, we'd get this error message:
```
python3: src/libumockdev-preload.c:81: get_libc_func: Assertion `fp' failed.
```

Update the code to throw an error on stderr before failing, for example:
```
umockdev: could not get libc function __xstat64
```


martinpitt (approved):
Nice, thank you! I'll look at #108 ASAP, and set up CI for rawhide.

PR #112, opened 2020-11-25
Don't display grep 'binary file matches' errors

seb128:
Those warnings a printed on stderr since the new grep 3.5 version,
which fails the Debian autopkgtest

From the grep NEWS
```
 Noteworthy changes in release 3.5 (2020-09-27) [stable]

** Changes in behavior

  The message that a binary file matches is now sent to standard error
  and the message has been reworded from "Binary file FOO matches" to
  "grep: FOO: binary file matches", to avoid confusion with ordinary
  output or when file names contain spaces and the like, and to be
  more consistent with other diagnostics.  For example, commands
  like 'grep PATTERN FILE | wc' no longer add 1 to the count of
  matching text lines due to the presence of the message.  Like other
  stderr messages, the message is now omitted if the --no-messages
  (-s) option is given.
```

### DIFF
--- a/tests/test-static-code
+++ b/tests/test-static-code
@@ -7,7 +7,7 @@ ROOTDIR=$(dirname $(readlink -f $0))/..
 CODE_INT=`grep -B 100000 'Overridden libc' $ROOTDIR/src/libumockdev-preload.c`
 CODE_WRAPPERS=`grep -A 100000 'Overridden libc' $ROOTDIR/src/libumockdev-preload.c`
 
-LIBC_OVERRIDES=`echo "$CODE_WRAPPERS" | grep '^[a-z_]\+(' | grep -v '^WRAP_' | cut -f1 -d'('`
+LIBC_OVERRIDES=`echo "$CODE_WRAPPERS" | grep --text '^[a-z_]\+(' | grep -v '^WRAP_' | cut -f1 -d'('`
 
 RET=0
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Hey @seb128 , many thanks for catching that! Ugh, I should have looked.. No tests were running here, apparently my Travis credits are used up :-( So I need to move my CI someplace else first.

All tests on Fedora and Ubuntu succeeded the last time, so maybe Debian does something special? Defaulting to a different locale or so?

Anyway, I'll sort out CI, then let's merge this ASAP, and then I'll see to enabling Debian tests on upstream PRs.

Merci beaucoup !

seb128:
> All tests on Fedora and Ubuntu succeeded the last time, so maybe Debian does something special? Defaulting to a different locale or so?

It's failing on Ubuntu as well [1] but it's not the tests which are not passing, it's just autopkgtest that disapproves of having stderr content, which maybe isn't an issue for your fedora tests?

I'm unsure why the previous version didn't trigger those with the new grep, but the output changed slightly displaying the summary of ok/expected fail/fail/unexpected pass/skipped/timeout as well now 

[1] https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-hirsute/hirsute/amd64/u/umockdev/20201118_035402_8ec3b@/log.gz

martinpitt:
@seb128: Oh! I see, yes, my upstream tests wouldn't catch that. I'll have a look at this.

Also, I don't really like the `-s` option here. Several calls to `grep` in that script already use the `--text` option, and I think it should be consistent.

seb128:
right, fair enough, I've changed -s to --text and force pushed

martinpitt:
@seb128: I just [fixed CI](https://github.com/martinpitt/umockdev/actions/runs/383278666), tests are running in GitHub now. Do you mind force-pushing again, so that they can run on the PR? Merci !

martinpitt (approved):
Cheers!

martinpitt:
I now added a test which ensures that installed-tests doesn't write anything to stderr. I [ran it on a branch](https://github.com/martinpitt/umockdev/runs/1453874430?check_suite_focus=true) without this fix, and it duefully reproduces the problem.

martinpitt:
.. and the tests rebased to master [pass now](https://github.com/martinpitt/umockdev/runs/1453956915).

PR #113, opened 2020-11-27
record: Do not filter out firmware_node

benzea:
I am trying to create some tests for thermald, and it is accessing the
firmware_node. As such, it is helpful to store the information in the
recording.


martinpitt (approved):
Sure, thanks!

Issue #96, opened 2019-09-20
umockdev-record does not record anything if recorded program overrides SIGINT


msergenergen:
I want to mock touchpad. But I didn't record anything into ioctl file.
When I was checking to mouse.ioctl file, I saw only this part
> @DEV /dev/input/event5

Probably, I didn't record anything. But I didn't get any error messages after following step;
`sudo umockdev-record -i /dev/input/event5=mouse.ioctl -e /dev/input/event5=mouse.events -- evtest /dev/input/event5`

 But I got this error message after following step;
`umockdev-run -d mouse.umockdev -i /dev/input/event5=mouse.ioctl -e /dev/input/event5=mouse.events -- evtest /dev/input/event5`

> ERROR: libumockdev-preload: failed to load ioctl record file for /dev/input/event5: empty or invalid format

Do you want to recommend anything else?

msergenergen:
I tried running on another pc and devices (keyboard, external mouse, touchpad) (umockdev installation). But I didn't get any result.

msergenergen:
Also I tried install and run steps on another Ubuntu versions. (16,04 and 18,04) I got same error.

DanielMalangu:
Yes, I have this exact issue. I'm not sure why no one has attempted to provide a solution. Is this a problem that is not common at all?

msergenergen:
Probably many people did not occur this problem.

hadess:
You're missing one half of the commands, you also need to create the `.umockdev` file:
https://github.com/martinpitt/umockdev/blob/master/README.rst#record-and-replay-input-devices

(even though that doesn't work for me either)

hadess:
Nothing is getting saved to disk:
```
$ sudo umockdev-record /dev/input/event18 > lap-proximity-switch.umockdev
$ sudo UMOCKDEV_DEBUG=all umockdev-record -i /dev/input/event18=lap-proximity-switch.ioctl -e /dev/input/event18=lap-proximity-switch.events -- evtest /dev/input/event18 
testbed wrapped open(/dev/input/event18) -> /dev/input/event18
ioctl_record_open: Starting new record lap-proximity-switch.ioctl
ioctl_record_open: starting ioctl recording of fd 3 into lap-proximity-switch.ioctl
init_script_dev_logfile_map: will record script of device 13:82 into lap-proximity-switch.events
script_record_open: start recording fd 3 on device 13:82 into lap-proximity-switch.events (format 1)
[...]

PR #118, opened 2021-02-25
Server side ioctls

benzea:
This is somewhat work-in-progress. But, it is working now, so opening an MR as there is something that can be discussed.

I think that ideally we would share more code paths between recording and replay. i.e. use a umockdev testbed for recording purposes. But I have not looked into the implications that this may have for the testbed API.


martinpitt:
Thanks @benzea for your great work here! This is quite a mouthful, so I suggest to spoonfeed this in several smaller PRs. The first for the first five commits, which should be relatively easy. Then one for the commits related to thread → mainloop. And finally one for the ioctl.

FTR, ignore the ubuntu:devel test failure -- this is a [glibc bug](https://launchpad.net/bugs/1916485) which can't be worked around at the moment, so I disabled that test for now in commit 0aab5f82fa56 -- so if you rebase your smaller PRs, they should be green.

martinpitt (changes requested):
Review of the first few "simple" commits. Thanks!

tests/test-umockdev-record.vala
```diff
@@ -71,9 +71,9 @@ t_testbed_all_empty ()
 
     spawn ("umockdev-record" + " --all", out sout, out serr, out exit);
 
-    assert_cmpstr (serr, Op.EQ, "");
-    assert_cmpstr (sout, Op.EQ, "");
-    assert_cmpint (exit, Op.EQ, 0);
+    Assertions.assert_cmpstr (serr, Op.EQ, "");
```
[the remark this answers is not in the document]
martinpitt: I tried to drop umockdev's assertions.vapi, and with that I get tons of
```
../../source/tests/test-umockdev-run.vala:436.5-436.18: error: The name `assert_cmpuint' does not exist in the context of `t_gphoto_download'
    assert_cmpuint ((uint) st.st_size, Op.GT, 5000);
```

There still only seems to be [GLib.assert()](https://valadoc.org/glib-2.0/GLib.assert.html), and a few non-comparison ones like [GLib.assert_error()](https://valadoc.org/glib-2.0/GLib.assert_error.html). We could probably use the [standard Check vapi](https://valadoc.org/check/Check.assert_str_eq.html), but last time I tried them they were not as nice as glib's.

src/umockdev.vala
```diff
@@ -1389,6 +1393,15 @@ public class Testbed: GLib.Object {
             return -1;
     }
 
+    private void worker_run ()
+    {
+        worker_ctx.push_thread_default();
+
+        worker_loop.run();
+
```
martinpitt: No need for the empty lines, that's a bit excessive.

src/umockdev.vala
```diff
@@ -75,6 +75,10 @@ public class Testbed: GLib.Object {
         this.dev_fd = new HashTable<string, int> (str_hash, str_equal);
         this.dev_script_runner = new HashTable<string, ScriptRunner> (str_hash, str_equal);
 
+        this.worker_ctx = new MainContext();
+        this.worker_loop = new MainLoop(this.worker_ctx);
+        this.worker_thread = new Thread<void> ("umockdev-testbed-worker", this.worker_run);
```
martinpitt: This makes sense, but I think this commit should be squashed into one of the next ones that actually *uses* the worker thread. Especially because the next commit rewrites/reverts most of the changes here.

src/umockdev-record.vala
```diff
@@ -417,7 +428,41 @@ main (string[] args)
     foreach (string s in opt_evemu_events)
         record_script(s, "evemu");
 
-    Posix.execvp(opt_devices[0], opt_devices);
-    exit_error("Cannot run program %s: %s", opt_devices[0], strerror(errno));
-    return 0;
+    int status;
+    try {
+        Process.spawn_async (null, opt_devices, null,
```
martinpitt, line 482: Nice! This should even address issue #96, as now the parent pid is always guaranteed to get a SIGINT and can cleanly close its recording, right?

src/umockdev-record.vala
```diff
@@ -417,7 +428,41 @@ main (string[] args)
...
+                            SpawnFlags.SEARCH_PATH | SpawnFlags.CHILD_INHERITS_STDIN | SpawnFlags.DO_NOT_REAP_CHILD ,
+                            null, out child_pid);
+    } catch (Error e) {
+            stderr.printf ("Cannot run %s: %s\n", opt_devices[0], e.message);
+            Process.exit (1);
+    }
+
+    // propagate signals to the child
+    var act = Posix.sigaction_t() { sa_handler = child_sig_handler, sa_flags = Posix.SA_RESETHAND };
```
martinpitt, line 492: This whole commit is basically a copy of how umockdev-run spawns its child. As this is far from trivial, let's try to factorize this.

src/umockdev-record.vala
```diff
@@ -444,6 +478,7 @@ main (string[] args)
                             null, out child_pid);
     } catch (Error e) {
             stderr.printf ("Cannot run %s: %s\n", opt_devices[0], e.message);
+            remove_dir (root_dir);
```
martinpitt, line 487: Is it really that important to clean up in this corner case? As C/glib makes it so excessively annoying to recursively remove a directory, let's just not bother?

src/umockdev-record.vala
```diff
@@ -430,6 +454,16 @@ main (string[] args)
...
 
+    try {
+        root_dir = DirUtils.make_tmp("umockdev.XXXXXX");
+        Environment.set_variable("UMOCKDEV_DIR", root_dir, true);
+    } catch (FileError e) {
+        error("Cannot create temporary directory: %s", e.message);
+    }
+
+    // "disable" the testbed, the library will still try to load the ioctl
+    FileStream.open(Path.build_filename(root_dir, "disabled"), "w");
```
martinpitt, line 471: This should be closed again.

tests/test-umockdev-record.vala
```diff
@@ -165,23 +165,6 @@ A: knobs/red=off
...
-    string serr;
-    int exit;
-
-    var tb = new UMockdev.Testbed ();
-    tb.add_devicev ("mem", "zero", null, {"dev", "1:5"}, {});
-    spawn ("umockdev-record" + " --ioctl /sys/devices/zero=/dev/stdout -- " + readbyte_path + " /dev/zero",
-           out sout, out serr, out exit);
-    Assertions.assert_cmpint (exit, Op.NE, 0);
-    Assertions.assert_cmpstr (sout, Op.EQ, "");
-    assert (serr.contains ("UMOCKDEV_DIR cannot be used"));
```
martinpitt, line 182: Please remove the corresponding check from the preload as well then. I actually wonder why that even works, as umockdev-record now creates/sets that env, so why doesn't ioctl_record_open() barf on this?

martinpitt (changes requested):
Some more review of the auxiliary commits.

tests/test-umockdev-record.vala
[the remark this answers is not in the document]
benzea: Hmm, it could be because I have a custom compiled glib inside jhuild.

I am seeing:
```
../tests/test-umockdev-record.vala:74.5-74.17: error: `assert_cmpstr' is an ambiguous reference between `GLib.assert_cmpstr' and `Assertions.assert_cmpstr'
    assert_cmpstr (serr, Op.EQ, "");
    ^^^^^^^^^^^^^
```

Maybe the collision is because for newer vala/glib it is exported in the `.vapi`?

src/umockdev-record.vala
benzea: Yep, that becomes a non-issue as ioctl recording is handled in the parent pid.

For scripts it'll still be the case for now though. As those have not been moved at this point.

src/umockdev-record.vala
benzea: Yeah. Ideally, I would love for all this to live inside `umockdev.vala`. i.e. make recording into a special operation mode of the testbed.

I have not figured out how to do that at this point though (didn't put too much thought into that so far though). So, maybe best to just have a small set of utilities elsewhere.

tests/test-umockdev-record.vala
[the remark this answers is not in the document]
benzea: So yeah, just checked and from vala 0.51.1 (there is no 0.51.0) onwards these functions are supported.

So, probably sane to just add a version check instead.

src/umockdev-record.vala
benzea: This happens implicitly as the returned `FILE*` is immediately out of scope.

tests/test-umockdev-record.vala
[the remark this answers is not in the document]
martinpitt: Nice! You just fixed this in #119, so resolving.

martinpitt:
@benzea : Want  to rebase this and look at the issues above, so that we don't let this bitrot too long? Thanks!

src/umockdev-record.vala
benzea: Eh, not a big deal either, considering we have the helper already?

src/umockdev-record.vala
benzea: OK, moved it all into a `umockdev-utils.vala` file, which gets compiled into a static library.

There is probably a bit more cleanup potential there now.

martinpitt:
The [CentOS 8 failure](https://github.com/martinpitt/umockdev/pull/118/checks?check_run_id=2678500764) looks like a too old Vala API (`The name `new' does not exist in the context of `GLib.Signal'`). Unless that's trivial to fix, I'd be happy to stop supporting CentOS 8 now, and move to CentOS 9 stream (once there is a container image).

The valgrind error on [Alpine](https://github.com/martinpitt/umockdev/pull/118/checks?check_run_id=2678500929) looks more serious though, some portability issue?

martinpitt:
I tested this branch interactively with evtest, on my "standard AT keyboard":

    sudo LD_LIBRARY_PATH=. UMOCKDEV_DEBUG=all ./umockdev-record -i /dev/input/event3=/tmp/i -e /dev/input/event3=/tmp/e -- evtest /dev/input/event3 

This works fine on master (despite issue #96 it does write a tree and evemu file), but it crashes immediately with this branch:
```
[...]
ioctl_simplestruct_init_from_bin: EVIOCGREP(80084503): size is 8 bytes
ioctl_tree_insert: node of type EVIOCGREP ptr 0x94c320 already exists
ioctl fd 3 request 80084503: emulated, result 0
    Repeat code 0 (REP_DELAY)
      Value    250
    Repeat code 1 (REP_PERIOD)
      Value     33
ioctl_simplestruct_init_from_bin: EVIOCGPROP(80F84509): size is 248 bytes
ioctl_tree_insert: node of type EVIOCGPROP ptr 0x94c320 already exists
ioctl fd 3 request 80F84509: emulated, result 8
Properties:
Testing ... (interrupt to exit)
ERROR: libumockdev-preload: emulation code requested invalid read from 0x1 + 4
ERROR: libumockdev-preload: Error communicating with ioctl socket, errno: 14
```

@benzea , does this work or fail for you?

benzea:
Ugh, ok …

So:
 * The Alpine failure was because `IOCTL_REQUEST_TYPE` is `int`, which I passed as `unsigned long` over the wire, and the integer got sign-extended, filling the upper bits with 0xffffffff. Fix is to just cast to `unsigned int`.
 * The crash you are seeing makes sense, because I just blindly trusted the encoded size in the `ioctl` number. We cannot do that, so added `ioctl_data_size_by_id` and updated the table to set the `real_size` to `0` if data should not be resolved.
 * About `GLib.Signal.new` I suppose we could create our own binding in our namespace for the relevant parts (`@new`, `emit_by_name`, `connect_object`). Not sure if that is worth it.

martinpitt:
Whee, nice work @benzea ! As I said, I'm happy to disable centos-8 CI at this point. RHEL 8.4 is out of the door, there is not terribly much demand for newer umockdev versions there IMHO. I'll disable it tomorrow, so ignore that test case now.

I'll give this another review tomorrow, but I'd actually like to merge this RSN before it becomes stale again, so that you can actually use it in fprint. (I don't want to divert your attention even longer :grin: ). Unless you still have some blocker on your list?

benzea:
Good stuff!

Yeah, it should be quite helpful for libfprint (in particular with the corresponding follow up MRs to add pcap USB replay and eventually also SPI replay hopefully).

benzea:
And no, no blocker from my side. Unless you see an issue with the exposed API at this point.

Note that for SPI I am planning to also allow overriding `write`/`read` (with a separate `handle-{write,read}` signals). Just in case that is relevant for (API) review.

src/libumockdev-preload.c
```diff
@@ -587,6 +587,7 @@ static void ioctl_record_sigint_handler(int signum)
     raise(signum);
 }
 
+#if 0
```
martinpitt: Let's drop the code instead of disabling it.

src/libumockdev-preload.c
```diff
@@ -612,60 +614,229 @@ record_ioctl(IOCTL_REQUEST_TYPE request, void *arg, int result)
...
+    fprintf(stderr, "ERROR: libumockdev-preload: Error communicating with ioctl socket, received EOF\n");
+    exit(1);
+
+con_err:
+    fprintf(stderr, "ERROR: libumockdev-preload: Error communicating with ioctl socket, errno: %d\n",
+	    errno);
+    exit(1);
+
+
+#if 0
```
martinpitt, line 633: dito

martinpitt (approved):
Thanks a lot! Just some trivialities, plus disabling the centos-8-stream test. I'll just push them to your PR if you don't mind.

benzea:
Of course, feel free to just force push into my branch.

benzea:
Wait, right, you already merged it! Thanks! :-)

PR #119, opened 2021-03-01
Cleanups from the server side ioctl MR

benzea:
Just the first commits that should be non-controversial.


tests/assertions.vapi
```diff
@@ -1,6 +1,7 @@
 [CCode (cprefix = "G", lower_case_cprefix = "g_", cheader_filename = "glib.h")]
 namespace Assertions {
-  public enum Op {
+#if !VALA_0_52
```
martinpitt, line 3: Feels like the whole file should be enclosed in this `#if`? (But I suppose it doesn't make much difference).

It's great that vala finally learned about the glib assertions! One less thing to carry around in the future.

martinpitt (approved):
Danke!

tests/assertions.vapi
benzea: That would be nice, but the files are doing a `using Assertions`, so the namespace needs to be defined :-)

PR #123, opened 2021-06-01
ioctl-tree: Fix USB submit URB data resolution

benzea:
Sorry! Turns out I did not test this enough …

----

Commit 95470c488bf0 ("ioctl-tree: Add function to query size of passed
structure") added a way to query the expected data size. Unfortunately,
we do not have an entry for URB submission, meaning that the data would
simply never be resolved.

Fix this by adding a new I_DUMMY entry without any functions that will
be ignored during execution.


martinpitt (approved):
Danke!

Issue #82, opened 2018-07-13
Replay USB control transfers?


superm1:
I was hoping to use umockdev to mock a USB device characteristics.  I normally interact with the device via HID reports.

I've found that trying to capture the interactions with both `--ioctl` and `--script` don't actually work though.

For the purpose of showing this, I wrote a really simple sample application and wrapper that could interact with two of the USB endpoints here:
https://github.com/superm1/umockdev-hid

I use it to try to pull out two bytes of data from the report embedded in the USB control transfer.

```
Running: umockdev-record /dev/bus/usb/001/035

Running: umockdev-record --ioctl=/dev/bus/usb/001/035=001.035.ioctl ./hidreport b06e
Querying 413c:b06e
Major: 1, Minor: 17

Running: umockdev-run --device=001.035.umockdev --ioctl=/dev/bus/usb/001/035=001.035.ioctl ./hidreport b06e
Querying 413c:b06e
Failed: failed to get report: control transfer failed: USB error on device 413c:b06e : Input/output error [-1]


Running: umockdev-record /dev/bus/usb/001/038

Running: umockdev-record --ioctl=/dev/bus/usb/001/038=001.038.ioctl ./hidreport b06f
Querying 413c:b06f
Major: 1, Minor: 5

Running: umockdev-run --device=001.038.umockdev --ioctl=/dev/bus/usb/001/038=001.038.ioctl ./hidreport b06f
Querying 413c:b06f
Failed: failed to get report: control transfer failed: USB error on device 413c:b06f : Input/output error [-1]
```

So my question is how to handle this situation?  Is there a proper way to capture those control transfers for replay?

superm1:
Also all of the sample data I've saved in a commit on that repo with the sample program.

superm1:
One more thing to add in case it helps to understand what's going on, here is what libusb is actually doing during those calls.

```
$ sudo LIBUSB_DEBUG=10 ./create_report.py 
libusb: debug [libusb_init] created default context
libusb: debug [libusb_init] libusb v1.0.21.11156
libusb: debug [find_usbfs_path] found usbfs at /dev/bus/usb
libusb: debug [op_init] bulk continuation flag supported
libusb: debug [op_init] zero length packet flag supported
libusb: debug [op_init] sysfs can relate devices
libusb: debug [op_init] sysfs has complete descriptors
libusb: debug [linux_udev_event_thread_main] udev event thread entering.
[...]

Issue #117, opened 2021-01-24
fingerprint readers don't work with umockdev-record ioctl trees 


martinpitt:
@benzea and @3v1n0, the [libfprint](https://gitlab.freedesktop.org/libfprint/libfprint) authors, currently have to employ some [hackish post-processing](https://gitlab.freedesktop.org/libfprint/libfprint/-/blob/cb332c/tests/convert-ioctl.py) ([docs](https://gitlab.freedesktop.org/libfprint/libfprint/-/blob/cb332c/tests/README.md#possible-issues)) of recorded ioctls to make them work. There is also some related [#fprint IRC conversation](https://www.pastery.net/zsaxcz/) about this.

Filing this issue to collect debugging information and some discussion about it.

Quoting the gist of the problem:

Often, drivers do a chain of commands like: A then B each with a different reply. However, `umockdev-record` could create a file like this:

    A
     reply 1
     reply 2
    B
     reply 1
     reply 2

In that case, records must be re-ordered:

    A
     reply 1
    B
     reply 1
    A
     reply 2
    B
     reply 2

martinpitt:
Without having delved into this yet (it's been many years since I wrote this code, and it was mostly experimental), there are two lines of investigation:

 * Check if the "next sibling" iteration can be salvaged to support this kind of "interleaved stateful" ioctls, where the request is identical, but the response is different. My first hunch is that the interleaving of A and B breaks the state keeping.
 * If that will break compatibility or is too involved, consider adding an option to umockdev-record like `--stateful` or `--linear` which will not do the grouping.

3v1n0:
FYI the IRC conversation is liked also here https://www.pastery.net/matzzg/ as the previous link is going to expire...

benzea:
> My first hunch is that the interleaving of A and B breaks the state keeping.

Yes, that would likely work. In a way it does feel wrong to me though as these devices have a more complex internal state. I would prefer just having things in-order and being able to throw an error immediately when something unexpected happens (rather than trying to recover and potentially causing infinite loops).

martinpitt:
Done in #124, thanks @benzea !

PR #124, opened 2021-06-01
Implement pcap based USB replay

benzea:
The ioctl-tree based replay is mostly useful for protocols that are fundamentally state-less. With state-full protocols, the ioctl-tree may randomly jump around, cause confusing situations and gets the order wrong during recording.

The pcap based replay is also able to play-back control transfers adding this feature to umockdev, even if outside of the original ioctl-tree implementation.
    
Note that umockdev does not support creating a recording. Recordings should be done using a separate tool such as wireshark.
    
Closes: #82, #117


.github/workflows/tests.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+      env: EXTRA_PACKAGES=libpcap-dev
```
martinpitt: Please add that directly to tests/run-alpine instead.

martinpitt (changes requested):
Any chance you could include a little pcap trace file and run the replay in a test, to ensure that this stays working, and actually works across OSes and architectures?

.github/workflows/tests.yml
benzea: Yeah, I realised that now, trying to fix that up currently :-)

benzea:
OK, I added a test. I just recorded plugging in a keyboard and hit some keys. Then did the minimal ioctls to grab two packets.

Not ideal, should exercise a large chunk of the relevant code, even if many corner cases re not covered.

src/umockdev-pcap.vala
```diff
@@ -0,0 +1,251 @@
+
```
martinpitt, line 1: Can you please add the standard copyright header, with your copyright?

tests/test-umockdev-vala.vala
```diff
@@ -737,6 +817,8 @@ main (string[] args)
   Test.add_func ("/umockdev-testbed-vala/usbfs_ioctl_tree_override_default_device", t_usbfs_ioctl_tree_override_default_device);
   Test.add_func ("/umockdev-testbed-vala/usbfs_ioctl_tree_xz", t_usbfs_ioctl_tree_xz);
 
+  Test.add_func ("/umockdev-testbed-vala/usbfs_ioctl_pcap", t_usbfs_ioctl_pcap);
```
martinpitt, line 820: Please squash the unit test into the previous commit that adds the feature.

martinpitt (changes requested):
This looks great, many thanks! Can you please still adjust README.md to mention that umockdev-run can load pcap files, and add an example how to generate them with wireshark and then load them?

Cheers!

README.rst
```diff
@@ -152,6 +165,46 @@ Command line: Record and replay PtP/MTP USB devices
...
+
+    Bus 001 Device 004: ID 06cb:00bd Synaptics, Inc. Prometheus MIS Touch Fingerprint Reader
+
+- Dump the sysfs device and udev properties:
+
+  ::
+
+    $ umockdev-record /dev/bus/usb/001/004 > fingerprint.umockdev
+
+- Use `wireshark` to record the bus in question (in this case bus 0). By default
```
martinpitt, line 188: Did you mean "bus 001" here, or is that another kind of "bus"?

benzea:
Hmm, sec, I am getting errors from libusb. I think I need to change an error code somewhere.

benzea:
OK, good now. I was returning ENOENT rather than EAGAIN if there was no reapable URB.

martinpitt (approved):
Looks great to me, thank you!

PR #126, opened 2021-06-02
tests: Increase socket stream test timeout

benzea:
umockdev accept's the socket stream connection before setting up the
thread that handles the connection and runs the script. Because of this,
there is a major delay, causing occasional test failures in the CI.

Increase the time we wait for the first message. It might be nicer to
move the accept() call into the thread that handles the script, but it
isn't really that important and requires some restructuring.


benzea:
Well … obviously this doesn't work. I had hoped to get rid of exactly that failure here.

benzea:
Going to reopen it for the correct patch.

PR #127, opened 2021-06-02
umockdev: Only start socket server thread after creating FDs

benzea:
Otherwise we may end up selecting an FD that does not exist, detecting
EOF and the thread quits immediately. This causes random failures of the
test suite.

A sample of a failure caused by this can be seen in https://github.com/martinpitt/umockdev/runs/2730392818?check_suite_focus=true#step:3:2025

FD is 0 here, but should obviously be non-zero.

EDIT: Update link; for some reason it ended up pointing to the later run.

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -1760,7 +1760,6 @@ private class SocketServer {
         this.running = true;
         this.socket_scriptfile = new HashTable<string, string> (str_hash, str_equal);
         this.script_runners = new HashTable<string, ScriptRunner> (str_hash, str_equal);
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Argh, what a stupid blunder.. This bug has been around since 2013, many thanks for fixing!

martinpitt:
@benzea FYI, I'm AFK until next Monday, in case you have more fixes queued.

benzea:
Nope, nothing else for now. I am hoping to get the SPI stuff working, but that will need some other people poking it some more.

PR #129, opened 2021-06-16
Fix discard and print messages on stuck replay

benzea:
This fixes discard (which I had apparently not tested really).

Another thing is that it adds some debug print if replay is stuck unexpectedly. That shouldn't happen normally, but could if the application has a large timeout and there is a buffer mismatch.

### DIFF
--- a/src/umockdev-pcap.vala
+++ b/src/umockdev-pcap.vala
@@ -44,6 +44,7 @@ internal class IoctlUsbPcapHandler : IoctlBase {
                                 USBDEVFS_CAP_ZERO_PACKET;
     private pcap.pcap rec;
     private Array<UrbInfo?> urbs;
+    private Array<UrbInfo?> discarded;
     private int bus;
     private int device;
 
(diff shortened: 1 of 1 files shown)


src/umockdev-pcap.vala
```diff
@@ -151,22 +151,57 @@ internal class IoctlUsbPcapHandler : IoctlBase {
...
 
             usb_header_mmapped *urb_hdr = (void*) cur_buf;
 
+            uint64 cur_pkt_time_ms = urb_hdr.ts_sec * 1000 + urb_hdr.ts_usec / 1000;
+
             /* Discard anything from a different bus/device */
             if (urb_hdr.bus_id != bus || urb_hdr.device_address != device)
                 continue;
 
+            /* Print out debug info, if we need 5s longer than the recording
```
martinpitt, line 196: 2s, but sure

martinpitt (approved):
Danke!

PR #131, opened 2021-06-17
pcap: Fix crash after reaching end of recording

benzea:
When the end of recording is reached, we might try to access URB
information that did not exist. Simply return immediately instead of
going on to fix this.

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -558,6 +558,8 @@ ioctl_emulate(int fd, IOCTL_REQUEST_TYPE request, void *arg)
 		errno = req.arg2;
 
 		IOCTL_UNLOCK;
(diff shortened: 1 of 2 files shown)


benzea:
So, there is an issue :-/

It seems that for libfprint (via libgusb), libusb is effectively busy looping to reap URBs. Now, I am not sure whether this is a problem because we are not notifying its poll correctly, or because emulation is being weird. I'll at least need to find a workaround here, I think.

benzea:
OK, I have a workaround … i.e. insert a `usleep(0)` to force the scheduler to run and give other threads a chance while the lock is released.

martinpitt (approved):
Ugh, that's some higher-order magic with the sleep(0), thanks for figuring this out!

PR #132, opened 2021-06-24
Public ioctl api

benzea:
On top of #125 (sorry, but need fixes from there to make sense).

API is not optimal yet:
 * We could remove the main context from the `IoctlBase` initialisation, then pass it when attaching the device
 * Then we can hide it completely (or at least provide a good default)
 * detach requires passing the handler again; we could just hide that behind a hash table for lookup

But, works nicely.

### DIFF
--- a/src/umockdev-ioctl.vala
+++ b/src/umockdev-ioctl.vala
@@ -144,38 +144,17 @@ public class IoctlData {
      * between two separate ioctl's).
      * It is very unlikely that such an explicit reload is needed.
      *
-     * Note pointers modified cannot be reloaded. As such, you may need to
-     * resolve the data again and it is usually a good idea to do so.
-     *
-     * The function will try to recursively update any resolved pointers. If
-     * the pointer was modified by the client it will not be resolved
-     * afterwards!
+     * Doing this unresolves any resolved pointers. Take care to re-resolve
+     * them and use the newly resolved #IoctlData in case you need to access
+     * the data.
      *
      * Returns: #TRUE on success, #FALSE otherwise
      */
     public bool reload() throws IOError {
         load_data();
 
-        IoctlData[] old_children = children;
-        size_t[] old_offsets = children_offset;
-
         children.resize(0);
         children_offset.resize(0);
-
-        for (int i = 0; i < old_children.length; i++) {
-            /* If the pointer has an unexpected value, then just drop the reference. */
-            if (*((size_t*) &data[old_offsets[i]]) != old_children[i].client_addr)
-                continue;
-
-            /* Reload and set correct pointer. */
-            old_children[i].reload();
-
-            *((void**) &data[old_offsets[i]]) = old_children[i].data;
-
-            /* Store in our new children array. */
-            children += old_children[i];
-            children_offset += old_offsets[i];
-        }
 
         return true;
     }
--- a/src/umockdev-ioctl.vala
+++ b/src/umockdev-ioctl.vala
@@ -259,6 +259,7 @@ public class IoctlData {
 public class IoctlClient : GLib.Object {
     private IoctlBase handler;
     private IOStream stream;
+    private GLib.MainContext _ctx;
     public string devnode { get; }
 
     public ulong request { get; }
@@ -330,8 +331,6 @@ public class IoctlClient : GLib.Object {
      * This call is thread-safe.
      */
     public void complete(long result, int errno_) {
-        IdleSource source;
-
         /* Nullify some of the request information */
         assert(_cmd != 0);
         _cmd = 0;
@@ -341,9 +340,7 @@ public class IoctlClient : GLib.Object {
         this.result_errno = errno_;
 
         /* Push us into the correct main context. */
-        source = new IdleSource();
-        source.set_callback(complete_idle);
-        source.attach(handler.ctx);
+        _ctx.invoke(complete_idle);
     }
 
     /**
@@ -526,7 +523,7 @@ public class IoctlClient : GLib.Object {
     private void notify_closed_cb()
     {
         /* Force into correct thread. */
-        handler.ctx.invoke(notify_closed_idle);
+        _ctx.invoke(notify_closed_idle);
     }
 
     internal IoctlClient(IoctlBase handler, IOStream stream, string devnode)
@@ -534,6 +531,7 @@ public class IoctlClient : GLib.Object {
         this.handler = handler;
         this.stream = stream;
         this._devnode = devnode;
+        this._ctx = GLib.MainContext.get_thread_default();
 
         /* FIXME: There must be a better way to do this in vala? */
         GLib.Signal.connect_object(this.stream, "notify::closed", (GLib.Callback) notify_closed_cb, this, GLib.ConnectFlags.SWAPPED);
@@ -573,7 +571,6 @@ private class StartListenClosure {
 }
 
 public class IoctlBase: GLib.Object {
-    public MainContext ctx { get; set construct; }
     private HashTable<string,Cancellable> listeners;
 
     static construct {
@@ -583,17 +580,7 @@ public class IoctlBase: GLib.Object {
     }
 
     construct {
-        /* XXX: Should we use the default or thread_default? */
-        if (ctx == null)
-            ctx = GLib.MainContext.@default();
         listeners = new HashTable<string,Cancellable>(str_hash, str_equal);
-    }
-
-    public IoctlBase(MainContext? ctx)
-    {
-        /* Executed after construct, so don't overwrite the default if not set. */
-        if (ctx != null)
-            _ctx = ctx;
     }
 
     ~IoctlBase()
@@ -628,7 +615,7 @@ public class IoctlBase: GLib.Object {
           listeners.remove(devnode);
     }
 
-    internal void register_path(string devnode, string sockpath)
+    internal void register_path(GLib.MainContext? ctx, string devnode, string sockpath)
     {
         assert(DirUtils.create_with_parents(Path.get_dirname(sockpath), 0755) == 0);
 
@@ -697,9 +684,9 @@ internal class IoctlTreeHandler : IoctlBase {
 
     private IoctlTree.Tree tree;
 
-    public IoctlTreeHandler(MainContext? ctx, string file)
+    public IoctlTreeHandler(string file)
     {
-        base (ctx);
+        base ();
 
         Posix.FILE f = Posix.FILE.open(file, "r");
         tree = new IoctlTree.Tree(f);
@@ -803,11 +790,11 @@ internal class IoctlTreeRecorder : IoctlBase {
     string device;
     private IoctlTree.Tree tree;
 
-    public IoctlTreeRecorder(MainContext? ctx, string device, string file)
+    public IoctlTreeRecorder(string device, string file)
     {
         string existing_device_path = null;
         Posix.FILE log;
-        base (ctx);
+        base ();
 
         this.logfile = file;
         this.device = device;
--- a/src/umockdev-ioctl.vala
+++ b/src/umockdev-ioctl.vala
@@ -31,13 +31,16 @@ internal bool signal_accumulator_true_handled(GLib.SignalInvocationHint ihint,
  * of the user provided structure.
  */
 public class IoctlData {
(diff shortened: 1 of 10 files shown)


martinpitt:
This can be rebased now.

benzea:
OK, did the API changes that I mentioned.

I think this is good to go now. The only downside is that it means we commit to the whole API on how data is resolved. But, considering it works well enough for the current usecases, I think that is reasonable.

src/umockdev-ioctl.vala
```diff
@@ -174,6 +197,17 @@ public class IoctlData {
         }
     }
 
+    /**
+     * umockdev_ioctl_data_dirty:
+     * @self: A #UMockdevIoctlData
+     * @recursive: Whether to recursively mark children as dirty
```
martinpitt, line 203: What are "children" in this context? Can you please explain this in the paragraph below?

src/umockdev-ioctl.vala
```diff
@@ -535,13 +617,91 @@ public class IoctlClient : GLib.Object {
...
+ * @handler: A #UMockdevIoctlBase
+ * @client: A #UMockdevIoctlClient
+ *
+ * Called when an ioctl is requested by the client.
+ *
+ * Access the #UMockdevIoctlClient::arg property of @client to retrieve the
+ * argument of the ioctl. This is a pointer sized buffer initially with the
+ * original argument passed to the ioctl. If this is pointing to a struct, use
+ * umockdev_ioctl_data_resolve() to retrieve the underlying memory and update
+ * the pointer. Resolve any futher pointers in the structure in the same way.
```
martinpitt, line 655: "fu*r*ther"

src/umockdev-ioctl.vala
```diff
@@ -535,13 +617,91 @@ public class IoctlClient : GLib.Object {
...
+ *
+ * After resolving the memory, you can access it as if it was local. Do not
+ * forget to call umockdev_ioctl_data_dirty() if you modify the memory.
```
martinpitt, line 658: Would there be much harm in always syncing back the memory in `client_complete()`, instead of tracking it with dirty explicitly? It's not like this ioctl emulation would be anywhere near "fast" :grin: 

I.e. calling `resolve` could set an internal flag that the `arg` is a pointer, and `complete` would then do the syncing.

src/umockdev-ioctl.vala
```diff
@@ -535,13 +617,91 @@ public class IoctlClient : GLib.Object {
...
+ */
+/**
+ * UMockdevIoctlBase::handle-read:
+ * @handler: A #UMockdevIoctlBase
+ * @client: A #UMockdevIoctlClient
+ *
+ * Called when a read is requested by the client.
+ *
+ * The result buffer is represented by #UMockdevIoctlClient::arg of @client.
+ * Retrieve its length to find out the requested read lenght. This is an empty
```
martinpitt, line 681: "lenght" typo

src/umockdev-ioctl.vala
```diff
@@ -535,13 +617,91 @@ public class IoctlClient : GLib.Object {
...
+ */
+/**
+ * UMockdevIoctlBase::handle-write:
+ * @handler: A #UMockdevIoctlBase
+ * @client: A #UMockdevIoctlClient
+ *
+ * Called when a write is requested by the client.
+ *
+ * The written buffer is represented by #UMockdevIoctlClient::arg of @client.
+ * Retrieve its length to find out the requested write lenght. The content of
```
martinpitt, line 698: same "lenght" typo

martinpitt (changes requested):
I gave this some more scrutiny, as it adds public API. I like this, it's very generic and flexible, great work! :clap:

src/umockdev.vala
```diff
@@ -745,6 +751,53 @@ public class Testbed: GLib.Object {
...
+     * @error: return location for a GError, or %NULL
+     *
+     * Attach an #UMockdevIoctlBase object to handle ioctl, read and write
+     * syscalls on the given device file. This allows emulating arbitrary
+     * ioctl's and read/write calls for which umockdev does not have builtin
+     * support.
+     *
+     * Returns: %TRUE on success, %FALSE on error.
+     */
+    public bool attach_ioctl (string? dev, IoctlBase handler) throws GLib.Error
```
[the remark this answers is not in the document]
benzea: No, it doesn't :-)

This was just a copy-paste error from the `.ioctl` loader, which will use the devnode from the `@DEV` header if not given.

If we want to let people override the catch-all handler, then we could add API for that later on. But, I don't feel that is useful.

src/umockdev.vala
```diff
@@ -745,6 +751,53 @@ public class Testbed: GLib.Object {
...
+     * Detach a custom #UMockdevIoctlBase object from an device node again. See
+     * umockdev_testbed_attach_ioctl(). It is an error to call this function
+     * if the path is not currently attached.
+     *
+     * Returns: %TRUE on success, %FALSE on error.
+     */
+    public bool detach_ioctl (string? dev) throws GLib.Error
+    {
+        IoctlBase handler = this.custom_handlers[dev];
+        handler.unregister_path(dev);
```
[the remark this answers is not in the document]
benzea: Changing it to `throw new IOError.NOT_FOUND("No handler for device was registered.");`

tests/test-umockdev-vala.vala
```diff
@@ -855,6 +855,86 @@ t_mt_uevent ()
...
+    client.complete(client.arg.data.length, 0);
+    return true;
+}
+
+static bool
+ioctl_custom_handle_read_cb(UMockdev.IoctlBase handler, UMockdev.IoctlClient client)
+{
+    UMockdev.IoctlData written = client.get_data("written");
+
+    assert(written != null);
```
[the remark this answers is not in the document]
benzea: Any real code should subclass `IoctlBase` and keep its data around that way.

The main issue would be keeping data per-connection (i.e. `IoctlClient`). We don't have a convenient way to do so other than using `g_object_set_data` right now or keeping around a hash table which really isn't better.

I can add some sanity checks and comments just in case someone gets the idea of copying this code. Ideally they would be copying the SPI or pcap code though.

src/umockdev-ioctl.vala
```diff
@@ -2,6 +2,16 @@
...
 namespace UMockdev {
 
+/**
+ * SECTION:umockdev-ioctl
+ * @title: umockdev-ioctl
+ * @short_description: Emulate ioctl and read/write calls for devices.
+ *
+ * These classes permit emulation of ioctl and read/write calls including
+ * fully customizing the behaviour by creating an #UMockdevIoctlBase instance or
+ * subclass instance and attaching it using umockdev_testbed_attach_ioctl().
```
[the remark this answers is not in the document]
benzea: It is already clickable. Functions are referenced by adding the `()`.

src/umockdev-ioctl.vala
benzea: We could do, it generates a few more context switches in cases where it isn't needed. Though, if we want to avoid that we could even create a backing buffer for the original client side data and check if anything has actually changed.

So, fine with me to remove this optimization and simplify the API that way.

martinpitt: API-wise it would be more robust, but admittedly I can't judge how bad it would affect performance, and if that matters at all. Mostly a question, I'll let you decide.

benzea: There are two different ways we can optimize this. i.e. `memcpy` and changing the protocol to avoid some of the round-trips. So, I think the right thing here is to make it implicit and remove it from the API.

This will also mean loading the read buffer from the application. But, not really a big concern.

At the end, I am mostly concerned about having unnecessary context switches. These seem to cost us a lot of time if there is CPU contention on the system (I suppose the scheduler is not fair with regard to overall CPU time in this case).

benzea: OK, did the changes. That does simplify the ioctl handling code quite a bit throughout to be honest.

The performance impact should be pretty small overall.

tests/test-umockdev-vala.vala
```diff
@@ -855,6 +855,86 @@ t_mt_uevent ()
...
+    assert(written.data.length == client.arg.data.length);
+
+    Posix.memcpy(client.arg.data, written.data, client.arg.data.length);
```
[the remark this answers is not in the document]
benzea: I changed it to return the start of the buffer. It still discards the full thing now, but added a comment to say that this will not usually make sense.

benzea:
OK, I think it is quite a bit nicer now from an API point of view.

src/umockdev.vala
```diff
@@ -745,6 +751,57 @@ public class Testbed: GLib.Object {
         this.ev_sender.send(devpath, action);
     }
 
+    /**
+     * umockdev_testbed_attach_ioctl:
+     * @self: A #UMockdevTestbed.
+     * @dev: (not nullable): Device path (/dev/...) to attach to.
```
martinpitt, line 757: TIL that `(not nullable)` is [really a thing](https://gi.readthedocs.io/en/latest/annotations/giannotations.html#type-signature), I've just never seen it. That's the default and should also be clear for valac due to the `string` type?

src/umockdev.vala
```diff
@@ -745,6 +751,57 @@ public class Testbed: GLib.Object {
...
+     * @dev: Device path (/dev/...) to detach.
+     * @error: return location for a GError, or %NULL
+     *
+     * Detach a custom #UMockdevIoctlBase object from an device node again. See
+     * umockdev_testbed_attach_ioctl(). It is an error to call this function
+     * if the path is not currently attached.
+     *
+     * Returns: %TRUE on success, %FALSE on error.
+     */
+    public bool detach_ioctl (string? dev) throws GLib.Error
```
martinpitt, line 794: For symmetry, and as the code does not expect NULL, please drop the `?` here as well.

tests/test-umockdev-vala.vala
```diff
@@ -855,6 +855,114 @@ t_mt_uevent ()
...
+{
+  var tb = new UMockdev.Testbed ();
+
+  tb_add_from_string (tb, """P: /devices/test
+N: test
+E: SUBSYSTEM=test
+""");
+
+  var handler = new UMockdev.IoctlBase();
+  var write_buf = new uint8[10] { 1, 2, 3, 4 ,5 ,6 ,7, 8, 9, 10 };
```
martinpitt, line 925: The comma/spaces are still wrong here. (That was already part of the previous round, but already marked resolved)

tests/test-umockdev-vala.vala
```diff
@@ -855,6 +855,114 @@ t_mt_uevent ()
...
+    }
+    return true;
+}
+
+static bool
+ioctl_custom_handle_write_cb(UMockdev.IoctlBase handler, UMockdev.IoctlClient client)
+{
+    /* NOTE:
+     *
+     * This code uses g_object_{get,set}_data for simplicty. Any realy code
```
martinpitt, line 882: simplic*i*ty and s/realy/real/

tests/test-umockdev-vala.vala
```diff
@@ -855,6 +855,114 @@ t_mt_uevent ()
   assert_cmpuint (change_count, CompareOperator.EQ, num_changes);
 }
 
+static bool
+ioctl_custom_handle_ioctl_cb(UMockdev.IoctlBase handler, UMockdev.IoctlClient client)
```
martinpitt, line 859: Can you please squash the API addition and this "Add test" commit?

martinpitt (changes requested):
Nice, thanks! It would be great to have a unit test that demonstrates how to handle ioctl with pointer resolving, but that can also come later -- it should already work as the SPI/evdev etc. tests use that functionality.

A few remaining leftovers only.

src/umockdev.vala
benzea: Oh, you are right! I thought I had double checked in the GIR that it was needed. But, turns out this happened due to the `?` and I actually forgot to drop the ? in `detach`.

benzea:
Thanks!

Some explicit testing for the IoctlData stuff would be nice for reference, yes. I did add a single `resolve` call to the existing test.

martinpitt (approved):
Yay, thank you! :clap:

PR #135, opened 2021-07-02
ioctl: Fix ioctl.h include location

benzea:
Including ioctl.h from asm-generic is not correct, as it will pick up
the wrong constants on some architectures. As such, include asm/ioctl.h
to avoid randomly overriding the values of some constants.

This fixes ppc64le.


martinpitt (approved):
Ugh, what a "fun" bug, thanks for tracking this down!

PR #136, opened 2021-07-07
Fix pcap replayer

benzea:
Stupid error in assert.

Also, looking at the code, I realised that control transfer playback cannot work right now. But, I am not sure what it should look like and don't have a good test case right now unfortunately (I hope to get some samples soon, if I do, it should be a quick fix).

### DIFF
--- a/src/ioctl.vapi
+++ b/src/ioctl.vapi
@@ -46,13 +46,13 @@ namespace Ioctl {
     [CCode (cheader_filename = "linux/usbdevice_fs.h")]
     public const int USBDEVFS_CAP_SUSPEND;
 
-    [CCode (cheader_filename = "asm/ioctl.h")]
+    [CCode (cheader_filename = "sys/ioctl.h")]
     public const int _IOC_SIZEBITS;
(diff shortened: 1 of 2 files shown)


martinpitt:
@benzea (you are aware of the conflict?)

src/ioctl.vapi
```diff
@@ -46,13 +46,13 @@ namespace Ioctl {
     [CCode (cheader_filename = "linux/usbdevice_fs.h")]
     public const int USBDEVFS_CAP_SUSPEND;
 
-    [CCode (cheader_filename = "asm-generic/ioctl.h")]
+    [CCode (cheader_filename = "asm/ioctl.h")]
```
martinpitt: This commit is already in master, I suppose that's what it complains about.

martinpitt (changes requested):

src/ioctl.vapi
benzea: Oopsa, no, I forgot to rebase!

martinpitt (approved):
Danke! Want/need a release with this?

PR #137, opened 2021-07-12
pcap: Implement control transfer replaying

benzea:
Control transfer replaying was not quite working. The primary reason was
that setup data needs to be compared differently from other buffer data.
Also, the code would incorrectly skip control transfers when there were
no reapable urbs.


martinpitt (approved):
Thanks! Argh special cases..

PR #143, opened 2021-09-08
pcap control fixes

benzea:
Issues discovered while creating the uru4000 test in libfprint.


martinpitt:
Tests timed out on ubuntu:devel, re-running.

martinpitt:
@benzea : Timed out again, this smells systematic..

benzea:
Yeah, though not a regression in umockdev.

The test that is timing out is the only one that is running Xorg internally. So, probably xorg is not shutting down.

src/umockdev-pcap.vala
```diff
@@ -256,8 +272,11 @@ internal class IoctlUsbPcapHandler : IoctlBase {
                         urb_buffer_length -= 8;
                     }
 
+                    /* libusb always sets URB_CONTROL endpoint to 0x00, but the
+                     * kernel exposes it as 0x80/0x00 depending on the direction
```
hadess, line 276: FWIW, the direction is a flag, the endpoint is only 4 bits:
https://www.beyondlogic.org/usbnutshell/usb6.shtml

benzea: Yeah. The funny thing is that `bmRequestType` and the endpoint byte of other packets are very similar yet different. So it would be correct to just copy the top bit from the setup data into the endpoint, but we can also just ignore it.

benzea:
OK, quite confusing. I added debug printf's, and we seem to be hanging inside the `spawn_async` call that launches `umockdev-run` and `Xorg`.

martinpitt:
Confirmed that the ubuntu:devel hang is not a regression from this PR, it just [failed the same way on master](https://github.com/martinpitt/umockdev/runs/3568593645?check_suite_focus=true). @benzea has a hunch that it's related to libglib2.0-0 2.68.3 → 2.68.4 update.

martinpitt (approved):
Thanks!

Issue #144, opened 2021-09-10
udev_enumerate_scan_devices() fails with ENOENT because of missing "bus" and "class" folders


dlech:
When using umockdev without loading any devices, `udev_enumerate_scan_devices()` will return `-ENOENT` since the "bus" and "class" directories have not been created yet. The [current test for this](https://github.com/martinpitt/umockdev/blob/6558285f957781dec03ddf0bd506873e68690e10/tests/test-umockdev.c#L128-L133) doesn't fail because [GUdev ignores the error from this function](https://gitlab.gnome.org/GNOME/libgudev/-/blob/master/gudev/gudevenumerator.c#L390).

PR #145, opened 2021-09-10
Testbed: Create "bus" and "class" dirs at init

dlech:
libudev expects the "/sys/bus" and "/sys/class" directories to exist when calling `udev_enumerate_scan_devices()`. When this was called on an empty `Testbed`, it would return `-ENOENT` since these directories did not exist.

This fixes the issue by creating the "/sys/bus" and "/sys/class" at the same time that "/sys" is created in the `Testbed`.

`num_udev_devices()` is modified to call `udev_enumerate_scan_devices()` directly instead of using GUdev, which ignores the return value of `udev_enumerate_scan_devices()`. This makes `t_testbed_empty()` a valid test for the bug we are fixing.

Some vala tests also had to be modified to account for the change.

Fixes: https://github.com/martinpitt/umockdev/issues/144

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -87,6 +87,12 @@ public class Testbed: GLib.Object {
         this.sys_dir = Path.build_filename(this.root_dir, "sys");
         DirUtils.create(this.sys_dir, 0755);
 
+        /* Create "bus" and "class" directories to make libudev happy */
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Very nice, thanks for the thorough fix and test!

PR #149, opened 2021-09-14
Avoid "never used" warnings for internal API

martinpitt:
We don't want to expose/commit to the registration API, so keep these
functions internal. However, they cause build warnings and break using
`werror`; so add some defines to conditionally expose these methods, so
that they only become available when linking to umockdev-record or
building the client library.


martinpitt:
Rebased to pick up ubuntu:devel fix.

benzea (approved):
Yeah. Doesn't win a beauty contest, but gets the job done.

PR #150, opened 2021-09-19
Add ioctls necessary to record/replay hidraw devices

ueno:
This series adds HID* ioctls necessary for testing devices only accessible through [HIDRAW](https://www.kernel.org/doc/html/latest/hid/hidraw.html) devices. My main use-case is to test [FIDO2 CTAP](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html) devices without physical hardware.

I have tested it with a signing [scenario](https://gist.github.com/ueno/007a4df8e0d58c64febaea76cb50f7dd#sign-and-verify-a-challenge), with the `fido2-assert` command from [libfido2](https://github.com/Yubico/libfido2):
```console
$ LD_PRELOAD=$PWD/random.so _build/umockdev-record --ioctl=/dev/hidraw5=fido2.ioctl --script=/dev/hidraw5=fido2.script -- fido2-assert -G -i assert_param /dev/hidraw5
$ LD_PRELOAD=$PWD/random.so _build/umockdev-run --device fido2.umockdev --ioctl=/dev/hidraw5=fido2.ioctl --script=/dev/hidraw5=fido2.script -- fido2-assert -G -i assert_param /dev/hidraw5
```
The `random.so` is a hack to replace `getrandom` used to generate an 8-byte [nonce](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-hid-init):
```console
$ cat random.c
#include <stdlib.h>
#include <string.h>

ssize_t __attribute__ ((visibility ("protected")))
getrandom (void *buf, size_t buflen, unsigned int flags)
{
  memset (buf, 0x1, buflen);
  return (ssize_t)buflen;
}
$ gcc -fPIC -shared -o random.so random.c
```


martinpitt:
This fails on all but the very latest OSes, apparently some ioctls are very new? E.g. [Alpine](https://github.com/martinpitt/umockdev/pull/150/checks?check_run_id=3648306018) and [Debian testing](https://github.com/martinpitt/umockdev/pull/150/checks?check_run_id=3648305948) don't know about `HIDIOC[GS]{IN,OUT}PUT`, and [ubuntu stable](https://github.com/martinpitt/umockdev/pull/150/checks?check_run_id=3648305892) additionally does not know `HIDIOCGRAWUNIQ`.

I ran `grep -r HIDIOCGINPUT /usr/include`, it's not just in a different header. Can you please guard these with `#ifdef`s, similar to `EVIOCGMTSLOTS`?

ueno:
Thank you for looking into it. According to the documentation `HIDIOC[GS]{IN,OUT}PUT` look like a new addition in linux 5.11, while `HIDIOCGRAWUNIQ` is undocumented but added in 2019. I've added `#ifdef` guards for them.

martinpitt:
Thanks! You'll alsoneed ifdefs for the unit test. But I started the tests to have a precise list of TODOs :grin: Thanks for bearing with this, the joys of supporting several operating systems..

martinpitt (approved):
I was wrong, the unit tests are now happy -- they don't use the unknown ioctls. Thanks!

PR #156, opened 2021-12-11
packit: Test on more architectures

martinpitt:
Also only do a COPR build, we don't currently have FMF tests.


martinpitt:
This now covers all the native architectures in COPR. I'll enable the emulated s390x/armhfp next, as these are problematic.

PR #166, opened 2022-01-18
build: Check whether run_command() succeeded

hadess:
As per:
```
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300
```


martinpitt (approved):
Merci !

Issue #167, opened 2022-01-20
Property buffer overflow starting at 0.17.4


wmww:
Commit b781cf65d49dc18d6319fb783c5617a71134bbbd broke Mir's umockdev tests (https://github.com/MirServer/mir/issues/2285). We get this error:
```
ERROR: uevent_sender_send: Property buffer overflow
Aborted (core dumped)
```
In GDB, I tracked down the problem to calling `append_property()` with a `name` that is too long for the `size`. On my machine, the name is (note that snaps I have installed seem to be included, no idea why):
```
(gdb) p name
$2 = 0x555557184f80 "DEVNAME=dri/card0DEVTYPE=drm_minor\nMAJOR=226\nMINOR=0\nSUBSYSTEM=drm\nID_PATH=pci-0000:00:02.0\nID_PATH_TAG=pci-0000_00_02_0\nID_FOR_SEAT=drm-pci-0000_00_02_0\nDEVLINKS=/dev/dri/by-path/pci-0000:00:02.0-card\nTAGS=:uaccess:snap_glimpse-editor_glimpse-editor:master-of-seat:snap_qemu-virgil_arm:snap_snap-store_ubuntu-software-local-file:snap_chromium_chromedriver:snap_super-cool-app_super-cool-app:snap_fluffychat_fluffychat:snap_flokk-contacts_flokk-contacts:snap_krita_krita:snap_qemu-virgil_qemu-virgil:snap_chromium_chromium:seat:snap_qemu-virgil_arm64:snap_firefox_firefox:snap_vlc_vlc:snap_snap-store_snap-store:snap_snap-store_ubuntu-software:snap_google-play-music-desktop-player_google-play-music-desktop-player:snap_discord_discord:\nCURRENT_TAGS=:snap_flokk-contacts_flokk-contacts:snap_chromium_chromium:snap_discord_discord:uaccess:snap_super-cool-app_super-cool-app:snap_fluffychat_fluffychat:snap_vlc_vlc:snap_snap-store_snap-store:snap_qemu-virgil_arm64:snap_qemu-virgil_arm:master-of-seat:snap_google-play-music-desktop-player_google-play-music-desktop-player:snap_glimpse-editor_glimpse-editor:snap_chromium_chromedriver:seat:snap_qemu-virgil_qemu-virgil:snap_firefox_firefox:snap_snap-store_ubuntu-software:snap_krita_krita:snap_snap-store_ubuntu-software-local-file:\n"
```
[...]

PR #168, opened 2022-01-20
Increase uevent_sender_send() buffer size

wmww:
Hacky fix for #167

### DIFF
--- a/src/uevent_sender.c
+++ b/src/uevent_sender.c
@@ -226,7 +226,7 @@ append_property(char *array, size_t size, size_t offset, const char *name, const
 void
 uevent_sender_send(uevent_sender * sender, const char *devpath, const char *action, const char *properties)
 {
-    char buffer[1024];
+    char buffer[16384];
     size_t buffer_len = 0;
     struct msghdr smsg;
     struct iovec iov[2];


martinpitt:
This fails expectedly, as this now fails the test that checks for overflows:
```
# child process (/umockdev-testbed/uevent/action_overflow [1477]) stderr: "**\nERROR:../tests/test-umockdev.c:850:t_testbed_uevent_action_overflow: child process (/umockdev-testbed/uevent/action_overflow [1480]) did not fail as expected\n"
Bail out! ERROR:../tests/test-umockdev.c:851:t_testbed_uevent_action_overflow: stderr of child process (/umockdev-testbed/uevent/action_overflow [1477]) failed to match: *uevent_sender_send*Property buffer overflow*
```

If you need longer properties for Mir, I'm happy to adjust the size, of course. But (1) I'd like to first track down the buffer overflow and catch it in this test (as obviously it currently doesn't spot the error), and (2) adjust that test along with the bigger buffer so that it still makes sense.

martinpitt:
Replaced with #170. Thanks, and sorry for the trouble!

After this lands, and after I fix #169, I'll do a new release (today still).

PR #174, opened 2022-02-27
Fix uevent race condition in add_from_string()

### LINKED ISSUE #173: Adding device fills some attributes after sending the uevent
There is a problem affecting hotplug in libusb. It reads the device descriptors immediately after it receives the uevent. However, the descriptor is a binary attribute.

I was setting it by using `umockdev_testbed_add_from_string`, but the uevent is already send by `umockdev_testbed_add_devicev`.

My thinking to fix this is to add `umockdev_testbed_add_devicev2`, etc. which do *not* emit the uevent. Then the API user needs to do so manually, fixing the issue. This allows firing the uevent from `umockdev_testbed_add_from_string`, fixing the issue mostly (though even there we could add `umockdev_testbed_add_from_string2` to move the responsibility to the API user.

Not sure what to do, but, as it is now, one needs to have all USB devices configured before initializing libusb. Otherwise setup will always race with the hotplug monitoring thread that libusb is running.

martinpitt:
We were sending the `add` uevent before setting up the binary/link
attributes and setting up the device node, which causes race conditions
in consumers.

To fix that, introduce an internal `add_devicev_no_uevent()` method which does
everything from testbed_add_devicev() except sending the uevent. Use 
that new helper from both that function and add_from_string() after everything
is ready.

Fixes #173

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -394,41 +394,9 @@ public class Testbed: GLib.Object {
         this.set_property(devpath, name, "%x".printf(value));
     }
 
-    /**
-     * umockdev_testbed_add_devicev:
-     * @self: A #UMockdevTestbed.
-     * @subsystem: The subsystem name, e. g. "usb"
-     * @name: The device name; arbitrary, but needs to be unique within the testbed
-     * @parent: (allow-none): device path of the parent device. Use %NULL for a
-     *          top-level device.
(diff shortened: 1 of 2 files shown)


benzea (approved):
Cool, yep, this fixes #173 without any API addition, which is kind of nice.

Also, thanks for adding the documentation about the functions sending "add" uevents. It is good to make that more explicit!

Issue #176, opened 2022-04-07
Help with debugging of -D_FORTIFY_SOURCE=3


marxin:
I'm in the middle of the debugging session where I'm trying to use `-D_FORTIFY_SOURCE=3` with the upcoming GCC compiler.
I noticed that using the flags for `systemd` and then using `/lib64/libudev.so.xyz` library triggers errors in your project test-suite:

```
[   16s] >>> LD_LIBRARY_PATH=/home/abuild/rpmbuild/BUILD/umockdev-0.17.8/x86_64-suse-linux GI_TYPELIB_PATH=/home/abuild/rpmbuild/BUILD/umockdev-0.17.8/x86_64-suse-linux G_DEBUG=fatal-warnings,fatal-criticals,gc-friendly PATH=/home/abuild/rpmbuild/BUILD/umockdev-0.17.8/x86_64-suse-linux:/usr/local/bin:/usr/bin:/bin TOP_SRCDIR=/home/abuild/rpmbuild/BUILD/umockdev-0.17.8 /home/abuild/rpmbuild/BUILD/umockdev-0.17.8/src/umockdev-wrapper /home/abuild/rpmbuild/BUILD/umockdev-0.17.8/x86_64-suse-linux/test-umockdev-record
[   16s] ――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
[   16s] stdout:
[   16s] # random seed: R02Se409341646f807cd297687b9042afab9
[   16s] 1..17
[   16s] # Start of umockdev-record tests
[   16s] # GLib-DEBUG: setenv()/putenv() are not thread-safe and should not be used after threads are created
[   16s] # DEBUG: umockdev.vala:110: Created udev test bed /tmp/umockdev.ZI5SJ1
[   16s] # DEBUG: umockdev.vala:137: Removing test bed /tmp/umockdev.ZI5SJ1
[   16s] # GLib-DEBUG: unsetenv() is not thread-safe and should not be used after threads are created
[   16s] ok 1 /umockdev-record/testbed-all-empty
[   16s] # GLib-DEBUG: setenv()/putenv() are not thread-safe and should not be used after threads are created
[   16s] # DEBUG: umockdev.vala:110: Created udev test bed /tmp/umockdev.WI9SJ1
[   16s] # DEBUG: umockdev.vala:765: umockdev_testbed_uevent: lazily initializing uevent_sender
[...]

PR #177, opened 2022-04-08
Provide wrapper for fortified version of readlinkat.

marxin:
Implement wrapper that is used when -D_FORTIFY_SOURCE
is used.

Fixes: #176.

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -1516,6 +1516,13 @@ ssize_t readlinkat(int dirfd, const char *pathname, char *buf, size_t bufsiz)
     return r;
 }
 
+/* A readlinkat fortify wrapper that is used when -D_FORTIFY_SOURCE is used. */
+ssize_t __readlinkat_chk(int dirfd, const char *pathname, char *buf, size_t bufsiz, size_t buflen);
+ssize_t __readlinkat_chk(int dirfd, const char *pathname, char *buf, size_t bufsiz, size_t buflen)
+{
+    return readlinkat(dirfd, pathname, buf, bufsiz);
+}
+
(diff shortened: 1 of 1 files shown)


src/libumockdev-preload.c
```diff
@@ -1516,6 +1516,12 @@ ssize_t readlinkat(int dirfd, const char *pathname, char *buf, size_t bufsiz)
     return r;
 }
 
+/* A readlinkat fortify wrapper that is used when -D_FORTIFY_SOURCE is used. */
+ssize_t __readlinkat_chk(int dirfd, const char *pathname, char *buf, size_t bufsiz, size_t buflen)
```
benzea, line 1520: I guess we need the same treatment here as for `__realpath_chk` and add a definition.

```suggestion
ssize_t __readlinkat_chk(int dirfd, const char *pathname, char *buf, size_t bufsiz, size_t buflen);
ssize_t __readlinkat_chk(int dirfd, const char *pathname, char *buf, size_t bufsiz, size_t buflen)
```

martinpitt (approved):
Thank you! See [my comment on the issue](https://github.com/martinpitt/umockdev/issues/176#issuecomment-1094195168) for some details.

marxin:
Thanks for merging it. Can I expect a release any time soon?

martinpitt:
@marxin ! Sure, if you need it. [here's 0.17.9](https://github.com/martinpitt/umockdev/releases/tag/0.17.9). I'll push it to Fedoras as well, do you need it in Debian/Ubuntu?

marxin:
Thanks for it. I need it for openSUSE and I've just made a package update submit request.

PR #179, opened 2022-04-14
Call libc's `readlink` directly in get_rdev

peadar:
Without this, we call the local readlink function, which will go
through the "WRAP3" version of readlink.

It generally avoids having the UMOCKDEV_PATH doubly appended,
so its mostly harmless from that point of view, but when
'-fno-semantic-interposition' and _FORTIFY_SOURCE are enabled during
compilation,  there's the potential for the wrapper functions to get
complex enough not to be inlineable, even though they are marked as
always_inline, resulting in a compile error.


martinpitt (approved):
Thanks for catching this! Indeed it was never intended to do the path resolution recursively.

PR #183, opened 2022-05-16
Documentation fixes

benzea:
I noticed that the documentation was basically empty on F36. The first commit fixes this, the second just solves a few minor issues.

I suspect this was triggered by an upstream valac change. Quite possibly gtk-doc upstream should be patched to accept VALA_EXTERN out of the box. I did not report an issue about that though.


martinpitt (approved):
Oh, vielen Dank!

Let's wait for green and land. I'd like to do a fix to packit, and then do another release.

PR #184, opened 2022-05-19
packit: Drop obsolete fields

martinpitt:
- `specfile_path` searches the tree for a .spec, and thus should be
   able to find it in packaging/ by itself.
 - `files_to_sync` already defaults to syncing the packit YAML and the
   spec, but this was broken for a few days
   (https://github.com/packit/packit/issues/1525)

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -1,17 +1,11 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
-specfile_path: packaging/umockdev.spec
 upstream_project_url: https://github.com/martinpitt/umockdev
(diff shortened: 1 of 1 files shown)


martinpitt:
This verifies dropping `specfile_path`, as that applies to the PR COPR builds. It can't verify `files_to_sync`, I'm taking `#packit`'s word for that. I'll do a new release afterwards to push out the doc fix, in the worst case I'll have to do another point release and put it back.

martinpitt:
> Local spec files found: [PosixPath('packaging/umockdev.spec')]. Taking: packaging/umockdev.spec

:+1:

PR #186, opened 2022-05-27
Environment API

benzea:
So, this adds a function returning an `char**` (same as `g_get_environ`).

Not sure what makes sense though. In python returning a dictionary would be more convenient. So maybe we should just expose both types of utility functions. This and a `GHashtTable` return or so.

Or just hash table.

Note that I kind of tried to get things to work using an xattr instead. But, turns out, tmpfs does not allow setting arbitrary values. I did manage to get it to run (except for nix) by setting a special ACL on the file (`system.posix_acl_access` xattr), but that is just one bit of information, which could probable be done using a sticky bit already ...

Or we could touch a `.umockdev` file in every device node, and then `openat(fd, ".umockdev")` to decide on whether to override ...

In conclusion, lets just do this, everything else is complicated and with questionable benefits.

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -149,7 +149,23 @@ public class Testbed: GLib.Object {
     }
 
     /**
-     * umockdev_testbed_get_root_dir:
+     * umockdev_testbed_get_environ:
+     * @self: A #UMockdevTestbed.
+     *
+     * Return the environment as it should be set for child processes.
+     *
+     * Returns: List of environment variables and their values.
+     */
+   [CCode (array_length = false, array_null_terminated = true)]
+    public string[] get_environ()
+    {
+        return {
+            "UMOCKDEV_DIR=%s".printf(this.root_dir),
+        };
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


benzea:
I suppose, python usage would be:

```
env = os.environ.copy()
env.update(v.split('=', 1) for v in testbed.get_environ())
```

martinpitt:
Thanks! TBH I don't really like this, it's sort of giving up on the idea of umockdev. Tomorrow morning I have a few hours in a train and I'll see if it's possible to mock the filesystem stat'ing enough to pacify udev without setting this env variable. If that fails, we can still add this.

benzea:
@martinpitt, please have a look at my [benzea/misc](https://github.com/benzea/umockdev/tree/benzea/misc) branch then!

benzea:
Oh, so, I just realised that /proc/self/fd actually shows the path for a file descriptor. So, maybe that is sufficient and much easier to do it then. e.g.

```ls -l /proc/self/fd 1000</sys/class/power_supply/BAT0/```

PR #187, opened 2022-05-30
preload: Wrap fstatfs()

martinpitt:
Revert commit 4b1ab11c244a0a488bb. Merely setting
`$SYSTEMD_DEVICE_VERIFY_SYSFS` does not work for GI bindings. Instead,
wrap fstatfs() and mangle the type to SYSFS_MAGIC if the path points
into our mocked /sys. This satisfies udev.

Fixes https://github.com/systemd/systemd/issues/23499

PR #188, opened 2022-05-30
statfs() follow-ups

martinpitt:
https://github.com/martinpitt/umockdev/pull/187#issuecomment-1140963786

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -1431,35 +1431,48 @@ int statx(int dirfd, const char *pathname, int flags, unsigned mask, struct stat
     return r;
 }
 
+static bool is_dir_or_contained(const char *path, const char *dir, const char *subdir)
+{
+    if (!path || !dir)
+	return false;
+
+    const ssize_t subdir_len = strlen(subdir);
+    const size_t dir_len = strlen(dir);
+
+    return (dir_len + subdir_len <= strlen(path) &&
+	    strncmp(path, dir, dir_len) == 0 &&
+	    strncmp(path + dir_len, subdir, subdir_len) == 0 &&
+	    (path[dir_len + subdir_len] == '\0' || path[dir_len + subdir_len] == '/'));
+}
+
+static bool is_fd_in_mock(int fd, const char *subdir)
+{
+    static char fdpath[PATH_MAX];
+    static char linkpath[PATH_MAX];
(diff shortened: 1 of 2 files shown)


src/libumockdev-preload.c
```diff
@@ -1431,35 +1431,44 @@ int statx(int dirfd, const char *pathname, int flags, unsigned mask, struct stat
     return r;
 }
 
+static bool is_fd_in_mock(int fd, const char *subdir)
+{
+    static char fdpath[PATH_MAX];
+    static char linkpath[PATH_MAX];
+    snprintf(fdpath, sizeof fdpath, "/proc/self/fd/%i", fd);
+    int orig_errno = errno;
+    ssize_t linklen = readlink(fdpath, linkpath, sizeof linkpath);
```
benzea, line 1440: No need to use our own overridden `readlink` version here.

src/libumockdev-preload.c
```diff
@@ -1474,6 +1474,28 @@ int fstatfs ## suffix(int fd, struct statfs ## suffix *buf)	\
...
+    const char *p = trap_path(path);					\
+    if (p == NULL || p == path) {					\
+	r = _statfs ## suffix(path, buf);				\
+	TRAP_PATH_UNLOCK;						\
+	return r;							\
+    } \
+    DBG(DBG_PATH, "testbed wrapped statfs" #suffix "(%s) -> %s\n", path, p); \
+    r = _statfs ## suffix(p, buf);					\
+    TRAP_PATH_UNLOCK;							\
+    if (r == 0 && (strcmp(path, "/sys") == 0 || strncmp(path, "/sys/", 5) == 0))	\
```
benzea, line 1491: I am starting to think that a helper function like `is_dir_or_contained` or so would be useful. The strcmp/strncmp is not entirely obvious to me when reading it and it is needed in multiple places. It could be inlined so that the compiler can optimize away the `strlen()` call to be static.

benzea (approved):

src/libumockdev-preload.c
martinpitt: Well spotted, thanks! Fixed.

src/libumockdev-preload.c
martinpitt: Good idea, done. I rely on gcc to figure out the optimization :grinning:

benzea (approved):
lgtm!

PR #189, opened 2022-07-01
Determine version dynamically from git

martinpitt:
Ship a ./version file so that this also works when building from dist
tarballs. Add a little helper shell script to determine this.
    
This enables us to reduce the release process to merely pushing a signed
tag.

----

See https://github.com/mesonbuild/meson/issues/688 for some details. Thanks to @eli-schwartz for your help with this!

After that I can bring back commit b316c0d438c


martinpitt:
A-ha! Seems the [rpm build](https://download.copr.fedorainfracloud.org/results/packit/martinpitt-umockdev-189/fedora-36-x86_64/04589703-umockdev/builder-live.log.gz) is tricked by some `.git` directory in the source tree which has no tags. The [generated srpm](https://download.copr.fedorainfracloud.org/results/packit/martinpitt-umockdev-189/fedora-36-x86_64/04589703-umockdev/umockdev-0.17.13-1.20220701100149223748.pr189.10.gc48aa27.fc36.src.rpm)'s tarball does *not* have such a .git dir.  However, I can reproduce the failure with `rpmbuild -rb *.src.rpm` on that srpm.

```
❱❱❱ git log
commit 7806e5d0c4d39987e4694e0901ae53bfdc2e3669 (HEAD -> master)
Author: rpm-build <rpm-build>
Date:   Fri Jul 1 12:22:10 2022 +0200

    umockdev-0.17.13 base
```

Oh well, then let's try a more robust approach.

eli-schwartz:
Does it for some reason have a shallow clone?

martinpitt:
@eli-schwartz : You mean rpm-build? No, that's just some completely unrelated git tree; I figure it's building an empty Fedora dist-git rpm tree or something such. I already fixed that part, just some little [version number syntax issue](https://download.copr.fedorainfracloud.org/results/packit/martinpitt-umockdev-189/fedora-36-x86_64/04589758-umockdev/builder-live.log.gz) left.

martinpitt:
Ah yes, alpine does not have /bin/bash, and Ubuntu's /bin/sh is dash which does not understand the `${VER/from/to}` syntax. So sed it is..

eli-schwartz:
Ah, that probably makes more sense, yeah. Checking that the project sources itself have a .git directory seems like the correct approach.

eli-schwartz:
This is one of the things that frequently motivates people to write meson install scripts in python, which is available due to meson itself depending on it.

Of course, you could install bash on Alpine if you really want.

And the other typical reason to prefer python, that neither bash nor /bin/sh nor indeed sed are available on Windows, is not really relevant to this project. ;)

martinpitt:
Hooray, green at last!

martinpitt:
@eli-schwartz That's a good point actually -- I keep forgetting that meson is python, so that's guaranteed to be installed.

PR #191, opened 2022-07-13
ioctl: Make data object usable from python

benzea:
So, things are not usable from python. Thing still broken:
 * We need the old _ref/_unref functions for API compatibility if we use a GObject, not sure how to do that. Or can we use a boxed type?
 * The length of data is not tagged in the GIR

So, kinda works, but you can't actually read the memory. And, not sure if maybe a `ctypes.memmove` shouldn't be workable instead.

The bindings cannot deal with a struct object, as it is not a boxed type
and it'll try to ref/unref using g_object_ref. Probably this is
fixable in another way, but this works.

Also, the python bindings cannot modify the byte data directly, so add
an update method that is essentially a glorified memcpy.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -26,6 +26,8 @@ add_project_arguments(
   '-Wno-error=pragmas',
   language: 'c')
 
+add_project_arguments('--abi-stability', language: 'vala')
+
 # let's not clutter the code with too many #ifdefs
 if get_option('b_ndebug') == 'true'
   add_project_arguments('-Wno-error=unused-but-set-variable', '-Wno-error=unused-variable', language: 'c')
--- a/src/umockdev-ioctl.vala
+++ b/src/umockdev-ioctl.vala
@@ -43,7 +43,7 @@ internal bool signal_accumulator_true_handled(GLib.SignalInvocationHint ihint,
  *
  * Since: 0.16
  */
-public class IoctlData {
+public class IoctlData : GLib.Object {
     /* Local cache to check if data is dirty before flushing. This is both an
      * optimization as it avoids flushes, but is also necessary to avoid
      * writing into read-only memory.
@@ -63,6 +63,30 @@ public class IoctlData {
     internal IoctlData(IOStream stream)
     {
         this.stream = stream;
+    }
+
+    /**
+     * umockdev_ioctl_data_ref:
+     * @self: A #UMockdevIoctlData
+     *
+     * Deprecated, same as g_object_ref().
+     */
+    [CCode(cname="umockdev_ioctl_data_ref")]
+    public new IoctlData? compat_ref()
+    {
+        return (IoctlData?) this;
+    }
+
+    /**
+     * umockdev_ioctl_data_unref:
+     * @self: A #UMockdevIoctlData
+     *
+     * Deprecated, same as g_object_unref().
+     */
+    [CCode(cname="umockdev_ioctl_data_unref")]
+    public new void compat_unref()
+    {
+        this.unref();
     }
 
     /**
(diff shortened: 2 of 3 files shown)


martinpitt:
Marking as draft until this works. Thanks!

benzea:
This works now, I believe. Adding the `update`/`retrieve` methods isn't really elegant, but I couldn't convince vala to put the array lenght information into the structure. And even if, we still need something like the `update` method anyway.

meson.build
```diff
@@ -26,6 +26,8 @@ add_project_arguments(
   '-Wno-error=pragmas',
   language: 'c')
 
+add_project_arguments('--abi-stability', language: 'vala')
```
martinpitt, line 29: What does this do?

src/umockdev-ioctl.vala
```diff
@@ -65,6 +65,30 @@ public class IoctlData {
         this.stream = stream;
     }
 
+    /**
+     * umockdev_ioctl_data_ref:
+     * @self: A #UMockdevIoctlData
+     *
+     * Deprecated, same as g_object_ref().
```
martinpitt, line 72: Why do we need this? Introducing an already deprecated API feels a bit strange, and these are not being used from anywhere.

martinpitt (commented):
Do you have some example how to use this? I'm happy to help with turning that into a test in test-umockdev.py. Is that parallel to t_ioctl_custom() in test-umockdev-vala.vala? Otherwise I'm afraid this will just bitrot and get broken.

Danke!

src/umockdev-ioctl.vala
benzea: Because vala defined these for the non-GObject based type. With GObject as the base class, we obviously don't need a custom ref/unref anymore, but need to add the methods to remain ABI compatible.

The only existing user (outside of umockdev itself) is the libusb test.

benzea:
> Do you have some example how to use this? I'm happy to help with turning that into a test in test-umockdev.py. Is that parallel to t_ioctl_custom() in test-umockdev-vala.vala? Otherwise I'm afraid this will just bitrot and get broken.

Yeah, a short test would be good, and something parallel to `t_ioctl_custom` is the right thing, yes. I think we can do something really minimal that just uses the new `retrieve/update` functions.

Yeah, it is pretty simple. My "existing user" was looking again a bit into the thermald test setup that I was hacking on at one point. Where I have the following code:

```
...
        self.acpi_thermal_rel_handler = UMockdev.IoctlBase()
        self.acpi_thermal_rel_handler.connect("handle-ioctl", self.acpi_thermal_rel_ioctl)

        self.testbed.attach_ioctl('/dev/acpi_thermal_rel', self.acpi_thermal_rel_handler)
...

    def acpi_thermal_rel_ioctl(self, handler, client):
        request = client.get_request()

        if request in (ioctls.ACPI_THERMAL_GET_ART_COUNT, ioctls.ACPI_THERMAL_GET_ART_LEN, ioctls.ACPI_THERMAL_GET_ART):
            print('ART')
            if self.acpi_art[0] < 0:
                client.complete(-1, -self.acpi_art[0])
                return True

            arg = client.get_arg()

            if request == ioctls.ACPI_THERMAL_GET_ART_COUNT:
                i = struct.pack('l', self.acpi_art[0])
                data = arg.resolve(0, len(i));
                data.update(0, i)
            elif request == ioctls.ACPI_THERMAL_GET_ART_LEN:
                i = struct.pack('l', len(self.acpi_art[1]))
                data = arg.resolve(0, len(i));
                data.update(0, i)
            elif request == ioctls.ACPI_THERMAL_GET_ART:
                data = arg.resolve(0, len(self.acpi_art[1]))
                data.update(0, self.acpi_art[1])

            client.complete(0, 0)
            return True

        elif request in (ioctls.ACPI_THERMAL_GET_TRT_COUNT, ioctls.ACPI_THERMAL_GET_TRT_LEN, ioctls.ACPI_THERMAL_GET_TRT):
            if self.acpi_trt[0] < 0:
                client.complete(-1, -self.acpi_trt[0])
                return True

            arg = client.get_arg()

            if request == ioctls.ACPI_THERMAL_GET_TRT_COUNT:
                i = struct.pack('l', self.acpi_trt[0])
                data = arg.resolve(0, len(i));
                print('TRT COUNT user side is:', data.retrieve(0))
                data.update(0, i)
            elif request == ioctls.ACPI_THERMAL_GET_TRT_LEN:
                i = struct.pack('l', len(self.acpi_trt[1]))
                data = arg.resolve(0, len(i));
                data.update(0, i)
            elif request == ioctls.ACPI_THERMAL_GET_TRT:
                data = arg.resolve(0, len(self.acpi_trt[1]))
                data.update(0, self.acpi_trt[1])

            client.complete(0, 0)
            return True

        return False
```

Let me see if I can hack up something quickly

martinpitt (approved):
Very cool, thank you!

benzea:
> Very cool, thank you!

I am a bit disappointed at myself for not checking the python (or GIR) output properly before the ioctl stuff got merged ... :-)

benzea:
Hmm, should i merge this, or do you want to do so before the next time you make a release?

martinpitt:
Ah sorry -- I wasnt' sure if you are done with this. Sure!

benzea:
Nah, I think it is ready, and simply had forgotten about it again :-)

PR #192, opened 2022-07-13
preload: Disable signals before locking

benzea:
Having signals enabled while holding a lock is messy as certain
recursions may occur. As such, disable signals before taking the mutex
and release signals again afterwards.


martinpitt:
Thank you! This looks good, the ubuntu:devel unit test [started to fail on main](https://github.com/martinpitt/umockdev/actions/workflows/tests.yml) 6 days ago on its own, apparently the new git version broke stuff again. I'll look at that first, and then rebase your branch.

martinpitt (approved):

PR #202, opened 2022-11-24
spec: Be generous with slow test timeouts

martinpitt:
Many koji architectures are emulated, especially in Pagure/packit builds. Don't be too picky about the timing there, similar to what we do in Debian autopkgtest.


martinpitt:
See e.g. https://koji.fedoraproject.org/koji/taskinfo?taskID=94493664 or https://koji.fedoraproject.org/koji/taskinfo?taskID=94493654

PR #207, opened 2023-02-26
umockdev: Fix race with worker thread accessing $UMOCKDEV_DIR

### LINKED ISSUE #206: "Environment" related crashes
The upower test suite randomly crashes in `getenv()`: https://gitlab.freedesktop.org/upower/upower/-/issues/216

I think that one of the possible reasons is the code that creates the testbed in `src/umockdev.vala` first creates a worked thread, and then sets `UMOCKDEV_DIR`. If the worker thread reads from the environment while the main thread is still setting up the environment, that could lead to a crash.

I would recommend:
- starting the worker thread after the `checked_setenv()` call
- making a copy of the value of `UMOCKDEV_DIR` and passing it as data to the thread you're creating, so it has its own copy of the string and doesn't need to poke at the environment from another thread.

martinpitt:
Set the env variable before starting the worker thread, to ensure that it sees it.

Fixes #206

### DIFF
--- a/src/umockdev.vala
+++ b/src/umockdev.vala
@@ -109,6 +109,8 @@ public class Testbed: GLib.Object {
         this.dev_script_runner = new HashTable<string, ScriptRunner> (str_hash, str_equal);
         this.custom_handlers = new HashTable<string, IoctlBase> (str_hash, str_equal);
 
+        checked_setenv ("UMOCKDEV_DIR", this.root_dir);
+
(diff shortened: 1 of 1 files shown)


martinpitt:
I'm not 100% sure if that will actually fix the issue, but it willl at least not make matters worse. The failed fedora:latest is due to https://bugzilla.redhat.com/show_bug.cgi?id=2166960.

Issue #213, opened 2023-07-02
System properties getting used for mocked device


hadess:
Using `systemd-udev-253.5-1.fc38.x86_64`, current git main umockdev, libgudev and upower master branches

Running the `test_headset_detection` test in upower, I get a failure on my desktop system because the headset device it detects shows up as an `internal` audio device, instead of a headset. The problem doesn't happen in the CI.

The properties dump for the device shows:
```
CURRENT_TAGS = :seat:
DEVPATH = /devices/pci0000:00/0000:00:14.0/usb1/1-8/1-8:1.0/sound/card0
ID_BUS = pci
ID_FOR_SEAT = sound-pci-0000_00_1f_3
ID_ID = usb-Logitech_G935_Gaming_Headset-00
ID_MODEL = G935_Gaming_Headset
ID_MODEL_ENC = G935\x20Gaming\x20Headset
ID_MODEL_FROM_DATABASE = Cannon Lake PCH cAVS
ID_MODEL_ID = 0xa348
ID_PATH = pci-0000:00:1f.3
ID_PATH_TAG = pci-0000_00_1f_3
ID_PCI_CLASS_FROM_DATABASE = Multimedia controller
ID_PCI_SUBCLASS_FROM_DATABASE = Audio device
ID_REVISION = 0112
ID_SERIAL = Logitech_G935_Gaming_Headset
ID_TYPE = audio
ID_USB_DRIVER = snd-usb-audio
ID_USB_INTERFACES = :010100:010200:030000:
ID_USB_INTERFACE_NUM = 00
ID_USB_MODEL = G935_Gaming_Headset
ID_USB_MODEL_ENC = G935\x20Gaming\x20Headset
ID_USB_MODEL_ID = 0a87
ID_USB_REVISION = 0112
ID_USB_SERIAL = Logitech_G935_Gaming_Headset
ID_USB_TYPE = audio
ID_USB_VENDOR = Logitech
ID_USB_VENDOR_ENC = Logitech
ID_USB_VENDOR_ID = 046d
ID_VENDOR = Logitech
ID_VENDOR_ENC = Logitech
ID_VENDOR_FROM_DATABASE = Intel Corporation
ID_VENDOR_ID = 0x8086
NVME_HOST_IFACE = none
SOUND_FORM_FACTOR = internal
SOUND_INITIALIZED = 1
SUBSYSTEM = sound
TAGS = :seat:
USEC_INITIALIZED = 12076650
```

[...]

PR #214, opened 2023-07-03
preload: Don't read udev cache data from system

hadess:
Never read the udev cache data from the system, as this might make properties of system devices leak through our mocked udev, such as headsets showing up as internal devices.

Closes: https://github.com/martinpitt/umockdev/issues/213


martinpitt:
Thanks! The change makes sense, I'll just check if I can add a unit test. (Probaby difficult as there's not a lot of udev data on trivial devices which are available in a container, but let's see..)

martinpitt (approved):
Ah of course, there's no /run/udev in a container at all, so let's just land this. Thanks!

PR #224, opened 2024-01-30
preload: wrap fstatfs() and statfs() on musl

alyssais:
musl has both of these functions (though not the separate "64" variants that glibc), and since systemd 251, a wrapped fstatfs() is required for udevmock's own test suite to pass, unless SYSTEMD_DEVICE_VERIFY_SYSFS is set to 0.

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -1397,53 +1397,6 @@ WRAP_3ARGS(ssize_t, -1, readlink, char *, size_t);
 WRAP_4ARGS(ssize_t, -1, getxattr, const char*, void*, size_t);
 WRAP_4ARGS(ssize_t, -1, lgetxattr, const char*, void*, size_t);
 
-#ifdef __GLIBC__
-WRAP_VERSTAT(__x,);
-WRAP_VERSTAT(__x, 64);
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks! But why did you move that code? First of all, it makes it very hard to see what actually changed (I don't see the musl fix, for example), and second it's now in a wrong place. If you want to move the code for aesthetic reasons, please do that as a separate commit, and then one with just the musl fix. Cheers!

alyssais:
I didn't move it for aesthetic reasons, I moved it so it wasn't in the `#ifdef __GLIBC__`.  Because I moved most of the code that was previously in there out, the way GitHub renders the diff is the `#ifdef` moving rather than the code that was previously inside it, as that results in a smaller diff.

martinpitt:
I see, thanks. However, I still don't like ripping apart the fstatfs{,64} and statfs{,64} declarations like that, they should stay together -- with the 64 variants being ifdeff'ed for glibc. I sent PR #225 with what I had in mind, which passes tests and seems to work well. If you care at all about having your name in "Author:" instead of just the commit message, please feel free to grab that diff, massage it a little bit if you like, and push it here -- I'm totally fine with that :grin:

PR #225, opened 2024-01-31
preload: wrap fstatfs() and statfs() on musl

martinpitt:
musl has both of these functions (though not the separate "64" variants that glibc).

Thanks to Alyssa Ross @alyssais for finding this and the initial patch!

---

This is an alternative to #224 which doesn't rip apart the fstatfs{,64} and statfs{,64} declarations.

Closes #224


alyssais (approved):

martinpitt:
Ack, thanks @alyssais for reviewing. Landing that then.

PR #234, opened 2024-03-03
pcap: Add appropriate endianness conversions for USB pcap header

benzea:
The USB pcap header should always be little endian. As such, add the appropriate conversions where necessary to ensure that the data is interpreted correctly on big endian architectures.

Maybe this fixes the sparc SIGBUS problem


benzea:
Hmm, that amd64 ubuntu:devel failure does not look like a regression from this patch (failing after `umockdev-testbed/libc` so `/umockdev-testbed/uevent/libudev`). Looks a little bit odd to be honest …

benzea:
Uh, did the test previously pass on s390x? I am a bit confused that the urb is not reaped in https://download.copr.fedorainfracloud.org/results/packit/martinpitt-umockdev-234/fedora-rawhide-s390x/07099222-umockdev/builder-live.log.gz :-)

But, I think I would need to debug on an s390x machine. Just staring at the code I don't see anything wrong.

src/umockdev-pcap.vala
```diff
@@ -216,10 +216,10 @@ internal class IoctlUsbPcapHandler : IoctlBase {
```
benzea, line 216: Hm, there is the `assert(cur_hdr.caplen)`. I guess that also needs a `uint32.from_little_endian`, but I am not entirely sure right now. It isn't going to cause issues though and we just rely on the USB header later on anyway.

martinpitt:
> that amd64 ubuntu:devel failure does not look like a regression from this patch

I haven't seen it fail recently, but I retried and it passed.

> Uh, did the test previously pass on s390x?

Yes, it did. Everything has been green for a while except for nix, and that got fixed this morning. So this is definitively a regression. Note that s390x is big-endian (for that very reason I want it in PR tests).

benzea:
> Yes, it did. Everything has been green for a while except for nix, and that got fixed this morning. So this is definitively a regression. Note that s390x is big-endian (for that very reason I want it in PR tests).

That is … odd. It must be pure luck that it is working, but no idea right now that could happen (possibly because those are mostly control transfers). Anyway, later.

Issue #237, opened 2024-04-12
remove_device does not trigger "remove" uevent


bobhenz-jabil:
First of all... this package is awesome for testing!

I would like to request an enhancement though...

While `add_device()` triggers an "add" uevent. Currently `remove_device()` doesn't trigger a "remove" uevent. I find that I have to first call `testbed.uevent(syspath, "remove")` then call `testbed.remove_device(syspath)`. This doesn't feel quite right to me since I don't have to do that when I'm adding a device.

In addition, the comment for `remove_device()` says "Note that this will also remove all child devices". So it would be _really_ nice if a "remove" uevent was generated for the children as well when a device is removed. (Similar to what happens on hardware.)

Again, I appreciate your work here. Thank you.

martinpitt:
You are right, it should trigger a remove uevent, for symmetry and also to better model what happens on a real system. I suppose nobody complained so far as that's a lot less common than adding :grin: 

I'll try to work on this over the weekend, but if you want to give it a try, please do!

bobhenz-jabil:
I might be able to look at it next week, I think the difficulty will be in generating uevents for the children but I need to understand the code better to understand how to do that.

bobhenz-jabil:
Just to state what I see as the difficulties here (I might be wrong on some of this):
1. A device can be added with the name `foo/bar/baz`, or 3 devices could be added called `foo`, `bar`, `baz`. I think the resulting directory structure would be the same right?
2. The remove event needs to include the attributes and properties that the add event contained.

Therefore, I have a design proposal:
 A. A hidden file (e.g. `.umockdev`) is added into the directory of each added device. The existence of this file would mean a device was added at this level.
B.  When a parent device is removed, the directory structure of all subdirectories under it would be walked (recursively) and any directory found with a file of that name in it, would have a uevent('remove') call made for it.
C. I don't know if the attributes and properties are already stored in the directory of devices, but if not (or if it's easier) the contents of the `.umockdev` file could be the properties and attributes sent with the add-event.

@martinpitt: Does this sound like a good approach?

bobhenz-jabil:
Great! I will try to come up with a branch that implements this using the `uevent` file over the next week.

bobhenz-jabil:
Actually I was able to get some time to work on this today (https://github.com/martinpitt/umockdev/pull/238). I am not familiar with .vala so please excuse any non-valaic ways of doing things. The following _does_ seem to work though (under the minimal testing I ran).

[...]

PR #238, opened 2024-04-24
Fully remove USB children when the parent USB device is removed

bobhenz-jabil:
Suggested fix for https://github.com/martinpitt/umockdev/issues/237


src/umockdev.vala
```diff
@@ -687,6 +714,50 @@ public class Testbed: GLib.Object {
         remove_dir(real_path, true);
     }
 
+    private string? find_a_child(string syspath)
```
martinpitt, line 717: I find this too complicated, and repeating too much logic. It feels to me that this should be a single `remove_device()` that just recursively calls itself (depth-first!). If you need extra arguments, make it a remove_device_with_children() and then remove_device() becomes an one-liner which calls the former.

src/umockdev.vala
```diff
@@ -687,6 +714,50 @@ public class Testbed: GLib.Object {
...
+          }
+          if (FileUtils.test(path, FileTest.IS_REGULAR)) {
+            if (name == "uevent") {
+              return syspath;
+            }
+          }
+        }
+
+        // If we're here, then this path is not a device, but it might
+        // have a child device hidden deeper.
```
martinpitt, line 737: This needs to happen *first*, i.e. the search needs to be depth-first. You need to remove the children first before the parents. This doesn't matter much in terms of removing the files (as that'll happen at the very end through `remove_dir()`, but it'll produce the wrong (inverted) order of `remove` events.

martinpitt (changes requested):
Thanks! This also needs some unit tests, to make sure that the remove events happen, and in the expected order.

Please let me know if/when it starts to overwhelm you, then I can take over.

src/umockdev.vala
bobhenz-jabil: Sorry, I will re-work.

src/umockdev.vala
bobhenz-jabil: It actually didn't invert the order, but you are correct that it is confusing. I will rework it per your recommendation above.

src/umockdev.vala
```diff
@@ -655,6 +655,9 @@ public class Testbed: GLib.Object {
             return;
         }
 
+        // Create a uevent for the removal of this device.
+        uevent(syspath, "remove");
```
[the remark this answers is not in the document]
bobhenz-jabil: I think this is a good idea, however, when I try to move the uevent call to be _after_ the removal of the directories (even if the reading of the uevent file properties is split out into a call before the directories are removed), I get this error:
```
ERROR: uevent_sender_send: No such device /sys/devices/pci0000:00/0000:00:14.0/usb1/1-3
```

Which I believe is coming from this line of code: `this.ev_sender.send(devpath, action, properties);`

bobhenz-jabil:
@martinpitt: Do these code changes look better?

src/umockdev.vala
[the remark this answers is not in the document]
martinpitt: Ah, you are right. That's a bit difficult to avoid, it'd need to make a copy of the uevent file (properties) first. But it doesn't hurt too badly indeed, let's go with this for now.

martinpitt:
@bobhenz-jabil sorry for the late response, it's holiday season here. The changes look good, thanks! They still need to grow some unit tests. If you want to add them, please do, otherwise I'll work on them.

But first I need to unbreak the build on Fedora, the new gcc broke vala/glib hard. I've already spent several hours on a "proper" solution, but I suppose I'll go with a quick hack for now.

martinpitt:
I just fixed the last remaining build issue (#239), and took the liberty to rebase your branch, so that the tests can run again. Note that this is still blocked by adding unit tests for the remove event and child removal.

bobhenz-jabil:
@martinpitt : My apologies for the lengthy delay. I have finally gotten around to adding the following commits to this PR:

1. Tweak an existing "uevent" test as it was manually invoking `uevent()` after adding a device when `add_device()` will invoke `uevent()` internally anyway.
2. Add a test for removing a single device. (verify that `uevent()` was called)
3. Add a test for removing a parent device. (verify that `uevent()` was called for it and each of its children)

Please let me know what you think.

src/umockdev.vala
```diff
@@ -644,6 +644,10 @@ public class Testbed: GLib.Object {
             return;
         }
 
+        // Generate the removal event (must be called before the directory
+        // structure is removed).
+        uevent(syspath, "remove");
```
martinpitt, line 692: Please move this below the next paragraph which gets/checks the subsystem. When calling this on invalid devices, there shouldn't be a `remove` event.

tests/test-umockdev.py
```diff
@@ -179,6 +179,42 @@ def on_uevent(client, action, device, counters):
         mainloop.run()
         self.assertEqual(counter, [0, 0, 1, syspath, '0815', '1'])
 
+    def test_remove_usb_device(self):
+        '''testbed removing a single USB device'''
```
martinpitt, line 183: This is too much duplication.  You can just append the `self.testbed.remove_device()` call to the existing `test_uevent()` above, run the main loop again, and check the counters. This is already set up for handling "remove" and "change" in the handler and counter.

tests/test-umockdev.py
```diff
@@ -215,6 +215,55 @@ def on_uevent(client, action, device, counters):
...
+        mainloop.run()
+        self.assertIn(parent_syspath, counter)
+        self.assertIn(child1_syspath, counter)
+        self.assertIn(child2_syspath, counter)
+        self.assertEqual(counter[parent_syspath], [1, 0])
+        self.assertEqual(counter[child1_syspath], [1, 0])
+        self.assertEqual(counter[child2_syspath], [1, 0])
+
+        # Just remove the parent, the children should also be removed.
+        self.testbed.remove_device(parent_syspath)
```
martinpitt, line 257: Can you please add an assertion that the parent syspath is gone from the file system after that? That implies that the children are gone, too.

martinpitt (changes requested):
Thanks! I've been meaning to add tests, but there's always something more urgent coming in between, I'm sorry!

This is now not far from landing any more, just needs some cleanups. Please also squash the unit test commits into the corresponding feature commits. I also suggest to move the "Test that add_device automatically creates an "add" uevent" commit first, as that already works on current main.

I'm happy to do that myself too if you don't have time or are fed up with this :wink:

martinpitt:
@bobhenz-jabil please ignore the debian-testing failures. It's a meson regression: https://github.com/mesonbuild/meson/issues/13461

tests/test-umockdev.py
bobhenz-jabil: Done.

tests/test-umockdev.py
bobhenz-jabil: Done.

bobhenz-jabil:
@martinpitt : Fixed-up the PR per your suggestions (I believe). Let me know if I missed anything.

martinpitt (approved):
Excellent, many thanks! :star2:

PR #240, opened 2024-05-21
packit: Enable propose-downstream for c9s, enable CentOS 10 Stream

martinpitt:


martinpitt:
Meh, it [still does the thing](https://artifacts.dev.testing-farm.io/fcff45d9-c93d-4778-aa33-fda7b7650c46/) where it downloads F40 repositories and often fails. I'll re-disable the test run, but the build is still useful as it runs the unit tests.

martinpitt:
c10s tests blocked on https://issues.redhat.com/browse/TFT-2612

Issue #252, opened 2024-07-20
preload: race condition wrt. pthread_sigmask


barathrm:
Hia and thanks for this awesome project :D 

### Context

I have a process which has been behaving oddly wrt. signals, and apparently only when run through `umockdev-run`. Specifically, my process has two threads. One thread blocks a specific signal right after being spawned, which is not blocked in the parent thread. Then, a moment later, the signal is seemingly unblocked again without me having touched it in that thread. I've written a minimum reproduction and pushed a PR which seems to fix the problem for me, with the caveat that I'm new to the codebase.

### Reproduction

Minimal reproduction code is pasted below, built with `clang++ main.cpp -o main -std=c++17 -Werror -fno-omit-frame-pointer -O2 -g ` . Example output, showing a signal being mysteriously unblocked in the spawned thread (the thread ID is prepended to each log output):

```
$ umockdev-run -- ./main
448903 --- start new thread
448903 --- Write to file
448906 +++ myThread
448906 +++ Signal SIGRTMAX is blocked: 0
448906 +++ Blocking signal
448906 +++ Signal SIGRTMAX is blocked: 1
448906 +++ Write to file
448903 --- Write to file
448906 +++ Signal SIGRTMAX is blocked: 0
448906 +++ Signal is not blocked!
448903 --- Write to file
^[[Aterminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument
```

```
#include <cstring>
#include <fstream>
#include <iostream>
#include <thread>

#include <signal.h>
#include <unistd.h>

const char *FILE_ONE = "one";
const char *FILE_TWO = "two";

bool writeEntireFile(const std::string &path, const std::string &data)
{
    std::ofstream file(path, std::ios::binary);
    file << data;
    file.close();
    return !file.fail();
}

int signal_is_blocked(int sig)
{
    sigset_t ss;
    int r;

    r = pthread_sigmask(SIG_SETMASK, NULL, &ss);
    if (r != 0) {
        return -r;
    }

    return sigismember(&ss, sig);
}
void myThread()
{
    std::cout << gettid() << " +++ myThread" << std::endl;
    int is_blocked = signal_is_blocked(SIGRTMAX);
    std::cout << gettid() << " +++ Signal SIGRTMAX is blocked: " << is_blocked << std::endl;
    std::cout << gettid() << " +++ Blocking signal" << std::endl;

    sigset_t ss{};
    int ret = 0;
    ret |= sigemptyset(&ss);
[...]

PR #253, opened 2024-07-20
preload: fix sigmask block and restore race

barathrm:
Related to and discussed more at https://github.com/martinpitt/umockdev/issues/252

---

To avoid clobbering a threads signal mask with that of another thread, take locks before storing a thread's signal mask in case another thread has already stored its own signal mask in the trap_path_sig_restore variable, but has not yet restored it.

Without this change, the following could happen:

If umockdev-run is used to run a process which spawns multiple threads, and each of those threads writes to a file, and has specific signal mask requirements, the following may happen

- Thread A: 
    - store thread A's sigmask in trap_path_sig_restore and block all signals for thread A 
    - acquire trap_path_lock

- Thread B: 
    - store thread B's sigmask in trap_path_sig_restore and block all signals for thread B. Thread A's original signal mask is lost. 
    - wait on acquiring trap_path_lock

- Thread A: 
    - "restore" thread A's sigmask using trap_path_sig_restore, which now contains thread B's sigmask, thus clobbering thread A's original sigmask with that of thread B.

### DIFF
--- a/src/libumockdev-preload.c
+++ b/src/libumockdev-preload.c
@@ -166,13 +166,13 @@ pthread_mutex_t ioctl_lock = PTHREAD_MUTEX_INITIALIZER;
     do { \
         sigset_t sig_set; \
         sigfillset(&sig_set); \
-        pthread_sigmask(SIG_SETMASK, &sig_set, &trap_path_sig_restore); \
         pthread_mutex_lock (&trap_path_lock); \
+        pthread_sigmask(SIG_SETMASK, &sig_set, &trap_path_sig_restore); \
     } while (0)
 #define TRAP_PATH_UNLOCK \
     do { \
-        pthread_mutex_unlock (&trap_path_lock); \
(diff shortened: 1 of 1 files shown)


martinpitt:
Hello @barathrm ! Oh dear, you clearly win my "bug report and analysis of the year" trophy! :trophy: Thanks so much for tracking this down!

The fix is straightforward, as usual in these cases (where there's a ton of work in debugging and understanding the problem).

Unfortunately [meson is making a boo-boo](https://github.com/martinpitt/umockdev/actions/runs/10021321500/job/27714819215?pr=253) in Debian testing. That has nothing to do with your fix, it just broke independently. I'll investigate this, and then rebase/queue your PR after that fix.

Cheers!

martinpitt (approved):
I reported the meson bug here: https://github.com/mesonbuild/meson/issues/13461

Not sure how to work around this, I don't want to block on this, so I'll ignore the Debian-testing failure. It's tested on all the other OSes.

Thanks again!

PR #254, opened 2024-09-04
Handle spurious "connection refused" when sending a uevent

bobhenz-jabil:
This error was already being handled on the `connect` call right above these changes, but wasn't being handled on the call to `sendmsg`. Depending on the timing it was possible for the `connect` call to succeed while the `sendmsg` call failed with an ECONNREFUSED error leading to the test crashing from the abort call.


bobhenz-jabil:
@martinpitt: I'm back with another PR. Let me know if you have questions or want changes.

martinpitt (approved):
Thanks @bobhenz-jabil ! The duplication is a bit annoying, but C just sucks at error handling.. :sweat_smile: 

I retried the (hopefully spurious) [CentOS 10 failure](https://download.copr.fedorainfracloud.org/results/packit/martinpitt-umockdev-254/centos-stream-10-x86_64/07978788-umockdev/builder-live.log.gz). I don't see how `python3-dnf` would really disappear, this just sounds like a package mirror hiccup.

PR #256, opened 2024-10-30
 tests: umockdev-record: t_system_single: handle missing SELinux context on /dev/null

g7:
/sys/fs/selinux might exist, yet the checked path might have no
context defined.

Drop the check for the selinux pseudo-fs, and check the eventual
errno if lgetfilecon() fails.

On ENODATA (context doesn't exist, or the process can't access it),
ENOTSUP (extended attributes not supported/disabled) and ENOSYS
(lgetxattr syscall used by lgetfilecon() not available), ensure
__DEVCONTEXT is not there and continue.

On every other error, fail the test case as before.

### DIFF
--- a/tests/test-umockdev-record.vala
+++ b/tests/test-umockdev-record.vala
@@ -201,13 +201,25 @@ t_system_single ()
     assert_in("P: /devices/virtual/mem/null", sout);
     assert_in("E: DEVNAME=/dev/zero", sout);
 #if HAVE_SELINUX
-    // we may run on a system without SELinux
-    if (FileUtils.test("/sys/fs/selinux", FileTest.EXISTS)) {
-        string context;
-        assert_cmpint (Selinux.lgetfilecon ("/dev/null", out context), CompareOperator.GT, 0);
-        assert_in("E: __DEVCONTEXT=" + context + "\n", sout);
+    string context;
+    int res = Selinux.lgetfilecon ("/dev/null", out context);
+    if (res > 0) {
+        assert_in ("E: __DEVCONTEXT=" + context + "\n", sout);
+    } else if (res == -1 && (Posix.errno == Posix.ENODATA ||
+                             Posix.errno == Posix.ENOTSUP ||
+                             Posix.errno == Posix.ENOSYS)) {
+        // If SELinux is not available, or is available but
+        // there is no context defined for /dev/null,
+        // we should skip this check as there is
+        // no context recorded.
+        //
+        // ENODATA: context doesn't exist, or the process
+        //          can't access it
+        // ENOTSUP: extended attributes not supported/disabled
+        // ENOSYS:  lgetxattr syscall not available
+        assert (!sout.contains("E: __DEVCONTEXT"));
     } else {
-        assert(!sout.contains("E: __DEVCONTEXT"));
+        assert_cmpint (res, CompareOperator.GT, 0);
     }
 #endif
 }


tests/test-umockdev-record.vala
```diff
@@ -202,11 +202,17 @@ t_system_single ()
     assert_in("E: DEVNAME=/dev/zero", sout);
 #if HAVE_SELINUX
     // we may run on a system without SELinux
+    int res = -1;
     if (FileUtils.test("/sys/fs/selinux", FileTest.EXISTS)) {
         string context;
-        assert_cmpint (Selinux.lgetfilecon ("/dev/null", out context), CompareOperator.GT, 0);
-        assert_in("E: __DEVCONTEXT=" + context + "\n", sout);
-    } else {
+        res = Selinux.lgetfilecon ("/dev/null", out context);
```
martinpitt, line 208: Do you know in which situations this can fail?

If we do that, then I think the /sys/fs/selinux check is moot, and can be dropped.

tests/test-umockdev-record.vala
```diff
@@ -202,11 +202,17 @@ t_system_single ()
...
+
+        if (res > 0) {
+            assert_in("E: __DEVCONTEXT=" + context + "\n", sout);
+        }
```
martinpitt, line 212: Does it fail with something useful, like `ENOSYS` or so? that could be asserted in the failing case.

martinpitt (changes requested):
Thanks! I'd like to understand this a bit better first, though.

tests/test-umockdev-record.vala
g7: Yes, it fails with `ENODATA`. Would you prefer adding it here explicitly (and fail if errno is not `ENODATA`/`ENOTSUP`)? I think it might be enough to catch these corner cases :smile:

tests/test-umockdev-record.vala
g7: Thanks for taking a look at this Pull Request!

Looking at the manpage, and reporting here for convenience:

```
If the context does not exist, or the process has no access to this attribute, errno is set to ENODATA.

If extended attributes are not supported by the filesystem, or are dis- abled, errno is set to ENOTSUP.

The errors documented for the [stat](https://linux.die.net/man/2/stat)(2) system call are also applicable here.
```

In my case, indeed it fails with `ENODATA`. In this specific case there is SELinux loaded, set as `permissive` and without any policy installed/loaded. So I think this is what tricked the test case.

Indeed, if we agree in keeping the check for the `lgetfilecon` return value, I think the `/sys/fs/selinux` check can be dropped, which is similar as what umockdev-record is doing.

martinpitt: Ack, then let's drop it...

tests/test-umockdev-record.vala
martinpitt: ... and check that the error is one of these, i.e. we know that it's due to "no selinux", not due to some other stupid stuff like ENOENT. After all, all of these devices are emulated, so I'd like to make sure that failures are under tight control.

g7:
v2:
* Drop check for /sys/fs/selinux
* Explicitly check against ENODATA, ENOTSUP  and ENOSYS
* Update commit (and Pull Request) description reflecting these new changes

martinpitt (approved):
Thanks! The gentoo error is unrelated, it already [started a month ago](https://github.com/martinpitt/umockdev/actions). I'll look at that separately.

martinpitt:
@thesamesam @mattst88: Do you happen to know why [valac does not get installed any more](https://github.com/martinpitt/umockdev/actions/runs/11624155887/job/32372326306)? It used to be part of 
```
ACCEPT_KEYWORDS="~*" emerge dev-util/umockdev --with-test-deps --onlydeps --getbinpkg
```
in the [run-gentoo test script](https://github.com/martinpitt/umockdev/blob/main/tests/run-gentoo#L23), but that doesn't happen any more. This was introduced by @mattst88 in ed29c0107ead81c1e460fff3884 and basically unchanged since then, other than the binary package optimization in commit 6a1d57a80a511c.

I can hack around it:
```diff
--- tests/run-gentoo
+++ tests/run-gentoo
@@ -21,6 +21,8 @@ $RUNC run --interactive ${OPTS:-} \
 echo 'FEATURES="\${FEATURES} binpkg-request-signature"' >> /etc/portage/make.conf
 # install build dependencies
 ACCEPT_KEYWORDS="~*" emerge dev-util/umockdev --with-test-deps --onlydeps --getbinpkg
+# HACK: for some reason not part of build deps any more
+emerge --getbinpkg dev-lang/vala
 # install git, "meson dist" dependency
 emerge --getbinpkg dev-vcs/git
```
but that feels bad -- vala very much needs to be a build dependency of umockdev. Is that invocation wrong? Any help appreciated, thanks!

martinpitt:
Sorry, let's discuss that Gentoo issue in PR #257 instead.

PR #259, opened 2024-12-21
Add umockdev_testbed_wait_script(), add API to add scripts/evemu as strings

### LINKED ISSUE #247: generating input events
Currently it is possible to play back recorded events with `umockdev_testbed_load_evemu_events()`. It would be very useful to be able to generate input events directly from the code, similar to `umockdev_testbed_uevent()`. Maybe there is a way I'm not aware of? 

I have tried:
```
cmd = g_strdup_printf("evemu-event --sync /dev/input/event1 --type EV_SW --code 15 --value 1");
g_assert(g_spawn_command_line_async(cmd, NULL));
```
but the events are nowhere to be found. `evtest` does not pick them up when called like that:
```
cmd = g_strdup_printf("evtest /dev/input/event1");
g_assert(g_spawn_command_line_async(cmd, NULL));
```
Note that evtest run this way can see events submitted by `umockdev_testbed_load_evemu_events()`.

martinpitt:
Fixes #247

This is mostly in a "this train ride ends here" state, but it works, and isn't too horrible. It could still use some fine-tuning, documentation, etc. @mkemlogic you weren't too happy about the names, we can do some bikeshedding :wink: 

What do you think?

### DIFF
--- a/tests/test-umockdev.c
+++ b/tests/test-umockdev.c
@@ -2022,8 +2022,6 @@ t_testbed_replay_evemu_events(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 {
   gboolean success;
   GError *error = NULL;
-  int fd;
-  g_autofree char *tmppath = NULL;
   struct timeval tv_begin, tv_end;
   struct input_event ev;
   static const char* test_data = "E: 1234.500000 0000 0000 0\n"   /* SYN */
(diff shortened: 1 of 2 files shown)


mkemlogic:
That was quick. I'm not sure I will have time before Christmas, but I should find some time before this year ends. It will be fun playing with this.

martinpitt:
@mkemlogic No worries, please enjoy your holidays! (As am I..) There is no rush here, we can talk about it in January.

martinpitt:
https://github.com/martinpitt/umockdev/commit/a51691040b3429335038e52dada8c5299d98231c went green, but with a retry. Doing a no-change push to re-run the tests, there could be a new flake.

tests/test-umockdev.c
```diff
@@ -1980,20 +2023,13 @@ t_testbed_replay_evemu_events(UMockdevTestbedFixture * fixture, UNUSED_DATA)
...
 
-  /* write evemu events file */
-  fd = g_file_open_tmp("test_evemu.XXXXXX", &tmppath, &error);
-  g_assert_no_error(error);
-  g_assert_cmpint(write(fd, test_data, strlen(test_data)), ==, strlen(test_data));
-  close(fd);
-
   /* load it */
-  success = umockdev_testbed_load_evemu_events(fixture->testbed, "/dev/input/event1", tmppath, &error);
+  success = umockdev_testbed_load_evemu_events_from_string(fixture->testbed, "/dev/input/event1", test_data, &error);
```
mkemlogic, line 2027: Should I be able to call `umockdev_testbed_load_evemu_events_from_string()`more than once?
I'm getting `ERROR:../src/umockdev.vala:1086:umockdev_testbed_load_script_from_stream: assertion failed: (!this.dev_script_runner.contains (owned_dev))`

I think this error was the motivation for my hack.

martinpitt: Yes, you should be able to do that. I just pushed an update which asserts that this works. However, in between two load calls you have to call `umockdev_testbed_wait_script()` to ensure that the previous script is finished. That will flush the script from the `this.dev_script_runner` dictionary.

martinpitt: > update which asserts that this works

I.e. I added unit tests for that, I didn't change the code. If that doesn't work for you, can you please point me to example code that fails? Thanks!

mkemlogic:
I'm sorry, I simply forgot about the `umockdev_testbed_wait_script()`.
It works perfectly now. 
I really appreciate adding this feature!

martinpitt:
@mkemlogic Cool, thank you for testing! This is a nice new feature indeed! I'll do a release in the next days.

martinpitt:
@mkemlogic released: https://github.com/martinpitt/umockdev/releases/tag/0.19.0 packaging for Fedora/Debian/Ubuntu in the next days.

PR #261, opened 2024-12-28
Fix ioctl_tree_execute() ret type and initialization

hdeller:
The umockdev-vala testcase randomly fails on some platforms (hppa, sparc64, powerpc, see:[1][2][3]) because the ioctl() testcase checks the return value pointed to by "ret" in ioctl_tree_execute() which might have stale data as it wasn't initialized.

Fix it by properly initializing the return code.
Additionally force all code paths to provide a valid return code address.

Tested on hppa-linux and armhf-linux.

[1] https://buildd.debian.org/status/fetch.php?pkg=umockdev&arch=hppa&ver=0.19.0-1&stamp=1735373270&raw=0
[2] https://buildd.debian.org/status/fetch.php?pkg=umockdev&arch=powerpc&ver=0.19.0-1&stamp=1735375553&raw=0
[3] https://buildd.debian.org/status/fetch.php?pkg=umockdev&arch=sparc64&ver=0.19.0-1&stamp=1735380455&raw=0

### DIFF
--- a/src/ioctl_tree.c
+++ b/src/ioctl_tree.c
@@ -357,6 +357,10 @@ ioctl_tree_execute(ioctl_tree * tree, ioctl_tree * last, IOCTL_REQUEST_TYPE id, 
     ioctl_tree *i;
     int r, handled;
 
+    /* initialize return code */
+    assert(ret != NULL);
+    *ret = -1;
+
     DBG(DBG_IOCTL_TREE, "ioctl_tree_execute ioctl %X\n", (unsigned) id);
 
     t = ioctl_type_get_by_id(id);
@@ -369,8 +373,6 @@ ioctl_tree_execute(ioctl_tree * tree, ioctl_tree * last, IOCTL_REQUEST_TYPE id, 
 	DBG(DBG_IOCTL_TREE, "  ioctl_tree_execute: stateless\n");
 	if (t->execute(NULL, id, arg, &r))
 	    *ret = r;
-	else
-	    *ret = -1;
 	return last;
     }
 
--- a/tests/test-umockdev-vala.vala
+++ b/tests/test-umockdev-vala.vala
@@ -271,7 +271,7 @@ E: SUBSYSTEM=usb
   assert_cmpint (Posix.ioctl (fd, Ioctl.USBDEVFS_CONNECTINFO, ref ci), CompareOperator.EQ, -1);
   // usually ENOTTY, but seem to be EINVAL
   assert_cmpint (Posix.errno, CompareOperator.GE, 22);
-  errno = 0;
+  Posix.errno = 0;
 
   // unknown ioctls don't work on an emulated device
   assert_cmpint (Posix.ioctl (fd, Ioctl.TIOCSBRK, 0), CompareOperator.EQ, -1);
--- a/src/ioctl_tree.vapi
+++ b/src/ioctl_tree.vapi
@@ -10,7 +10,7 @@ namespace IoctlTree {
 
       [ReturnsModifiedPointer]
       public void insert(owned Tree node);
-      public void* execute(void* last, ulong id, void* addr, ref int ret);
+      public void* execute(void* last, ulong id, void* addr, out int ret);
       [CCode (instance_pos = -1)]
       public void write(Posix.FILE f);
 
--- a/src/umockdev-ioctl.vala
+++ b/src/umockdev-ioctl.vala
@@ -572,7 +572,7 @@ public class IoctlClient : GLib.Object {
             IoctlData? data = null;
             ulong size = IoctlTree.data_size_by_id(_request);
             ulong type = (_request >> Ioctl._IOC_TYPESHIFT) & ((1 << Ioctl._IOC_TYPEBITS) - 1);
-            int ret = -1;
+            int ret;
             int my_errno;
 
             try {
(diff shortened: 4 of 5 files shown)


src/ioctl_tree.c
```diff
@@ -357,6 +357,9 @@ ioctl_tree_execute(ioctl_tree * tree, ioctl_tree * last, IOCTL_REQUEST_TYPE id,
     ioctl_tree *i;
     int r, handled;
 
+    /* initialize return code */
+    *ret = 0;
```
martinpitt, line 361: This "succeed by default" semantics is bad, as it hides errors. src/umockdev-ioctl.vala initializes the `ret` value to -1, so this is also inconsistent with the current behaviour.

Let's initialize to -1 here,  and drop the `else` in line 376 instead.

Let's also add an assertion that `ret` is not NULL, to document that requirement and make errors clearer.

tests/test-umockdev-vala.vala
```diff
@@ -268,17 +268,14 @@ E: SUBSYSTEM=usb
 
   /* no ioctl tree loaded */
   var ci = Ioctl.usbdevfs_connectinfo();
-  assert_cmpint (Posix.ioctl (fd, Ioctl.USBDEVFS_CONNECTINFO, ref ci), CompareOperator.EQ, -1);
-  // usually ENOTTY, but seem to be EINVAL
-  assert_cmpint (Posix.errno, CompareOperator.GE, 22);
-  errno = 0;
+  assert_cmpint (Posix.ioctl (fd, Ioctl.USBDEVFS_CONNECTINFO, ref ci), CompareOperator.EQ, 0);
```
martinpitt, line 271: The -1 check was quite deliberate, see above.

tests/test-umockdev-vala.vala
```diff
@@ -268,17 +268,14 @@ E: SUBSYSTEM=usb
 
   /* no ioctl tree loaded */
   var ci = Ioctl.usbdevfs_connectinfo();
-  assert_cmpint (Posix.ioctl (fd, Ioctl.USBDEVFS_CONNECTINFO, ref ci), CompareOperator.EQ, -1);
-  // usually ENOTTY, but seem to be EINVAL
-  assert_cmpint (Posix.errno, CompareOperator.GE, 22);
-  errno = 0;
```
martinpitt, line 274: This should go back. This test runs the actual ioctl, and checking the error code gives some additional confidence that this happens. It is of course more robust to also reset errno to 0 before, just like in the next check.

tests/test-umockdev-vala.vala
```diff
@@ -268,17 +268,14 @@ E: SUBSYSTEM=usb
...
+  assert_cmpint (Posix.ioctl (fd, Ioctl.USBDEVFS_CONNECTINFO, ref ci), CompareOperator.EQ, 0);
+  Posix.errno = 0;
 
-  // unknown ioctls don't work on an emulated device
-  assert_cmpint (Posix.ioctl (fd, Ioctl.TIOCSBRK, 0), CompareOperator.EQ, -1);
-  assert_cmpint (Posix.errno, CompareOperator.EQ, Posix.ENOTTY);
```
martinpitt, line 278: dito

martinpitt (changes requested):
Thanks for spotting this! Do you want to fix it up yourself, or want me to take over?

src/ioctl_tree.c
martinpitt: ... and also drop the then-redundant initialization in umockdev-ioctl.vala

martinpitt:
Thanks @hdeller ! I fixed the points above, and also cleaned up the commit message a bit.

martinpitt:
Thanks @hdeller ! I fixed the points above, and also cleaned up the commit message a bit. Before I do a new release, let's discuss/sort out #262

hdeller:
problem still happens on hppa build:
# DEBUG: umockdev.vala:1414: parsing device description for /devices/mycam
# DEBUG: umockdev.vala:1492: creating device /devices/mycam (subsystem usb)
# DEBUG: umockdev.vala:1538: create_node_for_device: creating file device /tmp/umockdev.SLZ0Z2/dev/001
# DEBUG: umockdev.vala:832: umockdev_testbed_uevent: sending uevent add for device /sys/devices/mycam
**
ERROR:../tests/test-umockdev-vala.vala:284:t_usbfs_ioctl_static: assertion failed (ioctl (fd2, TIOCSBRK, 0) == 0): (-1 == 0)
not ok /umockdev-testbed-vala/usbfs_ioctl_static - ERROR:../tests/test-umockdev-vala.vala:284:t_usbfs_ioctl_static: assertion failed (ioctl (fd2, TIOCSBRK, 0) == 0): (-1 == 0)

hdeller:
With this patch testcase succeeds on hppa and s390x:

diff --git a/tests/test-umockdev-vala.vala b/tests/test-umockdev-vala.vala
index 0d4add0..9cd88e2 100644
--- a/tests/test-umockdev-vala.vala
+++ b/tests/test-umockdev-vala.vala
@@ -281,8 +281,8 @@ E: SUBSYSTEM=usb
   // unknown ioctls do work on non-emulated devices
   int fd2 = Posix.open ("/dev/tty", Posix.O_RDWR, 0);
   if (fd2 > 0) {
-      assert_cmpint (Posix.ioctl (fd2, Ioctl.TIOCSBRK, 0), CompareOperator.EQ, 0);
-      assert_cmpint (Posix.errno, CompareOperator.EQ, 0);
+      assert_cmpint (Posix.ioctl (fd2, Ioctl.TIOCSBRK, 0), CompareOperator.EQ, -1);
+      assert_cmpint (Posix.errno, CompareOperator.EQ, Posix.ENOTTY);
       Posix.close (fd2);
   }

PR #262, opened 2024-12-29
tests: Disable spidev on big-endian 32-bit platforms

hdeller:
The class IoctlSpiBase (see src/umockdev-spi.vala) has a comment that it only works on 64bit platforms.
Disable those tests on 32-bit platforms for now.

This fixes my tests on the (32-bit) hppa platform.

This patch is just to show the problem (and a possible workaround).
If you find another way to fix (or skip) this test I'm fine with that,

Thanks!
Helge

### DIFF
--- a/tests/test-umockdev-vala.vala
+++ b/tests/test-umockdev-vala.vala
@@ -682,6 +682,12 @@ t_usbfs_ioctl_pcap ()
 void
 t_spidev_ioctl ()
 {
+  // does not work on 32 bit big-endian (such as Debian hppa)
+  if (sizeof(long) != sizeof(int64) && BYTE_ORDER == ByteOrder.BIG_ENDIAN) {
+      stdout.printf ("[SKIP: SPI emulation does not work on 32 bit big-endian] ");
+      return;
+  }
+
   var tb = new UMockdev.Testbed ();
 
   string device;


martinpitt:
Curious, CI runs on Debian i386 and alpine i386, and passes there.  But indeed src/umockdev-spi.vala mentions "This only works on 64bit machines", perhaps @benzea still remembers why. Could we make them more specific to hppa? https://en.wikipedia.org/wiki/PA-RISC says hppa is 64 bit big-endian, which would explain the `ulong` incompatibility. On the other hand, tests also run on s390x which is 64 bit big-endian too. So something doesn't add up here. Are you *sure* that your specific version of (or Debian's definition of) hppa is 32 bit?

hdeller:
Debian/hppa is currently 32-bit big-endian only (the Linux kernel may be either 32- or 64-bit though - similiar to 32bit userspace on 64bit amd64 machine).
Compared to s390x, "unsigned" is then a 32-bit value, while on s390x it's a 64-bit value.
Maybe that's the difference?
Should be possible to test again on both platforms on the debian porterboxes.

martinpitt:
Thanks! I adjusted the test to skip on 32 bit BE only, and also be more verbose about that fact (similar to how other tests are skipped).

martinpitt (approved):
Tested on the Debian hppa porter box. `meson test` passes now (despite #263 not yet being addressed)

PR #274, opened 2025-08-01
preload: Wrap `__fstat64_time64()`

martinpitt:
This is specific to Debian armhf (Fedora does not do it that way).

----

Fixes [Debian build failure](https://buildd.debian.org/status/fetch.php?pkg=umockdev&arch=armhf&ver=0.19.2-1&stamp=1753901783&raw=0)


martinpitt:
> Fedora does not do it that way

Oh! packit.yaml has `fedora-development-armhfp`, so I've always assumed that I *had* 32 bit arm coverage. But that doesn't run any more and gets silently ignored. There's no 32 bit arm in https://api.testing-farm.io/v0.1/composes/public any more, so I'll drop it from packit config.

